PR #15, opened 2015-11-25
fix docstring typo in jwe.py

nyx:


simo5:
Thanks you, reviewed and pushed.

PR #20, opened 2016-05-01
Add function to allow public JWKSet export

puiterwijk:
This function allows the export of the public parts of the
JWKSet.

Signed-off-by: Patrick Uiterwijk <puiterwijk@redhat.com>

PR #21, opened 2016-05-02
Provide a better way to export public keys

simo5:
This makes the same call export just the public keys if private_key is
set to False.

In hindsight if we had this from the start the default would be to export
public keys and you'd have to actively pass a True value to export private
ones, but I do not want to break applications now.

Not only this API it is somewhat more natural (with the above caveat), it also
allows to do exactlyu the same for JWK Sets with substantially no code
changes.

Signed-off-by: Simo Sorce <simo@redhat.com>

### DIFF
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -320,8 +320,14 @@ class JWK(object):
                                               ' "key_ops" values specified at'
                                               ' the same time')
 
-    def export(self):
-        """Exports the key in the standard JSON format"""
+    def export(self, private_key=True):
+        """Exports the key in the standard JSON format.
+
+        :param private_key(bool): Whether to export the private key.
+                                  Defaults to True.
+        """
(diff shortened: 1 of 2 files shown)


simo5:
@puiterwijk What do you think about this PR to replace #20 ?

jwcrypto/jwk.py
```diff
@@ -320,16 +320,21 @@ def import_key(self, **kwargs):
                                               ' "key_ops" values specified at'
                                               ' the same time')
 
-    def export(self):
+    def export(self, private_key=True):
         """Exports the key in the standard JSON format"""
```
puiterwijk: Maybe add the private_key argument to the docstring?

puiterwijk:
Replacing #20 with this sounds fine to me.

simo5:
Good catch, added doc strings

PR #23, opened 2016-05-09
Implement JWT deserialization by keyset rather than key

puiterwijk:
This allows passing of a keyset, and JWT will try to find the correct
key from the kid value in the header, and then decrypt or verify with
the correct key.

### DIFF
--- a/jwcrypto/jwt.py
+++ b/jwcrypto/jwt.py
@@ -8,6 +8,7 @@ from six import string_types
 from jwcrypto.common import json_encode, json_decode
 from jwcrypto.jws import JWS
 from jwcrypto.jwe import JWE
+from jwcrypto.jwk import JWKSet
 
 
 # RFC 7519 - 4.1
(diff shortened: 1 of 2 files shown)


jwcrypto/jwt.py
```diff
@@ -130,10 +167,12 @@ def __init__(self, header=None, claims=None, jwt=None, key=None,
         :param check_claims: An optional dict of claims that must be
          present in the token, if the value is not None the claim must
          match exactly.
+        :param keyset: A (:class:`jwcrypt.jwk.JWTSet`) keyset containing
+         the key to deserialize the token.
```
simo5: This option should be incompatible with passin a Key, can you ad mutual exclusivity across ?

jwcrypto/jwt.py
```diff
@@ -418,6 +457,16 @@ def deserialize(self, jwt, key=None):
         # have a key
         self.token.deserialize(jwt, key)
 
+        if keyset is not None:
```
simo5: we should probably test that keyset is an actual JWKset and raie a proper error if not ?

puiterwijk: Added.

jwcrypto/jwt.py
puiterwijk: Done, both in documentation and by throwing a ValueError.

simo5:
Superseeded by #26

PR #24, opened 2016-05-09
Implement reimport of previously exported JWKSet

puiterwijk:

### DIFF
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -505,6 +505,20 @@ class JWKSet(set):
             keys.append(json_decode(jwk.export(private_keys)))
         return json_encode({'keys': keys})
 
+    def import_keyset(self, keyset):
+        try:
+            jwkset = json_decode(keyset)
+        except:
+            raise InvalidJWKValue()
+
+        if 'keys' not in jwkset:
+            raise InvalidJWKValue()
+
+        for jwk in jwkset['keys']:
+            self.add(JWK(**jwk))
+
+        return self
+
     def get_key(self, kid):
         """Gets a key from the set.
         :param kid: the 'kid' key identifier.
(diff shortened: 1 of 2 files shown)


jwcrypto/jwk.py
```diff
@@ -482,6 +482,23 @@ class JWKSet(set):
...
+            except:
+                raise InvalidJWKValue()
+
+            for jwk in keyset:
+                self.add(JWK(**jwk))
+
+        elif keyset is None:
+            super(JWKSet, self).__init__()
+        else:
+            super(JWKSet, self).__init__(keyset)
```
simo5: What will this do?
Seems to me we should handle only None or a string that can be json_decode()d, not sure what good it makes to pass another unspecified value to the set() init ?

puiterwijk: This will call the set() constructor that takes an optional iterable argument to initialize it with.
Example: set([1,2,3]).
I didn't want to break anything in the API: I don't know if anyone would be using this constructor.

simo5:
Well if we do not want to break the constructor, then I guess we'd have to take *args, **kwargs in input and then find out if a keyset attribute was specified and do one thing, or call super withy *args, **kwargs ?

puiterwijk:
Well, passing kwargs on to the set constructor is probably a terrible idea, since it will barf out if it gets any keyword arguments.
We could still use a keyword argument, and if that's not provided just pass *args on to set.
(if len(args) > 1 it will crash very hard, but at least that's the same as right now).

simo5:
I would just pass anything we get in `*args` and `**kwargs,` of course if the arguments are wrong things will go south, but that is part of the contract of the set() primitive, should they change it in future we'll still be fully compatible.

simo5:
Thinking more about the interface ... we can let alone the __init__ function and instad provide a class interace called import() aand you'd call it like this:

newset = JWKSet.import(keyset_string)

The advanntaage is that we do not mess with init and we have almost symmetric import/export interfaces, except that import creates a new object. Bonus if we can make import return a new object only when called on the class constructor, and simply reset the object if called on an instance of the class.

simo5:
actually `JWKSet().import()` where import() ends with `return self` is probably the best option.

PR #26, opened 2016-05-16
Implement JWT deserialization by keyset

simo5:
This allows passing of a keyset, and JWT will try to find the correct
key from the kid value in the header, and then decrypt or verify with
the correct key.

Signed-off-by: Simo Sorce <simo@redhat.com>

### DIFF
--- a/jwcrypto/jwt.py
+++ b/jwcrypto/jwt.py
@@ -8,6 +8,7 @@ from six import string_types
 from jwcrypto.common import json_encode, json_decode
 from jwcrypto.jws import JWS
 from jwcrypto.jwe import JWE
+from jwcrypto.jwk import JWK, JWKSet
 
 
 # RFC 7519 - 4.1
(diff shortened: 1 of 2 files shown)


simo5:
This PR extends and replaces #23

simo5:
@puiterwijk the above is just a slight variation on your code and retains your commit, can you take a look and see if it works for you ?

jwcrypto/jwt.py
```diff
@@ -131,9 +168,9 @@ def __init__(self, header=None, claims=None, jwt=None, key=None,
          present in the token, if the value is not None the claim must
          match exactly.
 
-        Note: either the header,claims or jwt,key parameters should be
-        provided as a deserialization operation (which occurs if the jwt
-        is provided will wipe any header os claim provided by setting
+        Note: either the header,claims or jwt,key or jwt,keyset parameters
```
puiterwijk: This still references the keyset parameter.

puiterwijk:
Other than documentation, this looks good to me.

jwcrypto/jwt.py
simo5: good catch

puiterwijk:
ack

Issue #31, opened 2016-05-30
JWKSet doesn't allow for other keys


npmccallum:
The RFC explicitly allows the use of other key/value pairs in the JWKSet object. However, the library doesn't allow this. JWKSet should probably just implement the dict interface instead of the set interface.

simo5:
JWKSet is a set because we want to preserve order and uniqueness of keys.
RFC7517 does allow other members besides keys, I will amend that.

simo5:
I am looking at using an OrderedDict as the "keys" subdict and intercepting __setitem__ there to insure only JWK objects as elements, if everything works out the main object will be just a normal dict derived object, and will accept arbitrary elements for keys that are not "keys".

PR #33, opened 2016-05-31
Turn JWKSet into a special dictionary

simo5:
The JWKSet object is now a class derived from dict, with a special element
'keys' that is forced to be a class derived by set which enforces only JWK
object instances as elements.

A new classmethod "from_json" is added to create a JWKset out of a json object
directly.

The interface is fully backward compatible with the existing APIs otherwise.

### DIFF
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -6,7 +6,11 @@ from cryptography.hazmat.primitives.asymmetric import rsa
 from cryptography.hazmat.primitives.asymmetric import ec
 from jwcrypto.common import base64url_decode, base64url_encode
 from jwcrypto.common import json_decode, json_encode
+
+from six import iteritems
+
 import os
+
 
 # RFC 7518 - 7.4
 JWKTypesRegistry = {'EC': 'Elliptic Curve',
@@ -477,11 +481,7 @@ class JWK(object):
             raise NotImplementedError
 
 
-class JWKSet(set):
-    """A set of JWK objects.
-
-    Inherits for the standard 'set' bultin type.
-    """
+class _jwkset(set):
 
     def add(self, elem):
         """Adds a JWK object to the set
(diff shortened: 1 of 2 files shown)


simo5:
@npmccallum @puiterwijk please check this out
Ref Issue #31

jwcrypto/jwk.py
```diff
@@ -494,18 +494,53 @@ def add(self, elem):
...
+
+class JWKSet(dict):
+    """A set of JWK objects.
+
+    Inherits from the standard 'dict' bultin type.
+    Creates a special key 'keys' that is of a type derived from 'set'
+    The 'keys' attribute accepts only :class:`jwcrypto.jwk.JWK` elements.
+    """
+    def __init__(self, *args, **kwargs):
+        dict.__init__(self)
```
npmccallum: User super()

jwcrypto/jwk.py
```diff
@@ -494,18 +494,53 @@ def add(self, elem):
...
+        dict.__setitem__(self, 'keys', _jwkset())
```
npmccallum: User super()

jwcrypto/jwk.py
```diff
@@ -494,18 +494,53 @@ def add(self, elem):
...
+        self.update(*args, **kwargs)
+
+    def __setitem__(self, key, val):
+        if key == 'keys':
+            self['keys'].add(val)
+        else:
+            dict.__setitem__(self, key, val)
```
npmccallum: Use super()

jwcrypto/jwk.py
```diff
@@ -494,18 +494,53 @@ def add(self, elem):
...
+
+    def update(self, *args, **kwargs):
+        for k, v in iteritems(dict(*args, **kwargs)):
+            self.__setitem__(k, v)
```
npmccallum: self[k] = v

jwcrypto/jwk.py
simo5: This is what I got when I tried earlier:
    super(dict, self).__setitem__('keys', _jwkset())
AttributeError: 'super' object has no attribute '__setitem__'

That's why I am not using super on these.

jwcrypto/jwk.py
simo5: using self.__setitem__() is intentional, as it will handle the 'keys' key in the special way

jwcrypto/jwk.py
puiterwijk: Right, because it would be `super(JWKSet, self).__init__` etc.

simo5: Yeah, switching between languages makes you make the stupidest mistakes, revised patch pushed

npmccallum:
LGTM

PR #35, opened 2016-07-01
Generate a JWK from a cryptography (pyca) key

simo5:
If the key in input is a known python crypography key type then a
corresponding JWK set is generated from that key.


npmccallum:
ACK

PR #38, opened 2016-07-14
Add the missing JWE Key wrapping algorithms

simo5:
Adds AESGCMKW, PBES2 and ECDH-ES algorithms with tests.

@npmccallum can you please take a look ?

### DIFF
--- a/jwcrypto/jwe.py
+++ b/jwcrypto/jwe.py
@@ -151,13 +151,13 @@ class _RSA(_RawKeyMgmt):
     def __init__(self, padfn):
         self.padfn = padfn
 
-    def check_key(self, key):
+    def _check_key(self, key):
         if key.key_type != 'RSA':
             raise InvalidJWEKeyType('RSA', key.key_type)
 
     # FIXME: get key size and insure > 2048 bits
     def wrap(self, key, keylen, cek):
-        self.check_key(key)
+        self._check_key(key)
         if not cek:
             cek = os.urandom(keylen)
         rk = key.get_op_key('encrypt')
@@ -165,7 +165,7 @@ class _RSA(_RawKeyMgmt):
         return (cek, ek)
 
     def unwrap(self, key, ek):
-        self.check_key(key)
+        self._check_key(key)
         rk = key.get_op_key('decrypt')
         cek = rk.decrypt(ek, self.padfn)
         return cek
@@ -177,7 +177,7 @@ class _AesKw(_RawKeyMgmt):
         self.backend = default_backend()
         self.keysize = keysize // 8
 
-    def get_key(self, key, op):
+    def _get_key(self, key, op):
         if key.key_type != 'oct':
             raise InvalidJWEKeyType('oct', key.key_type)
         rk = base64url_decode(key.get_op_key(op))
@@ -186,7 +186,7 @@ class _AesKw(_RawKeyMgmt):
         return rk
 
     def wrap(self, key, keylen, cek):
-        rk = self.get_key(key, 'encrypt')
+        rk = self._get_key(key, 'encrypt')
         if not cek:
             cek = os.urandom(keylen)
 
@@ -209,7 +209,7 @@ class _AesKw(_RawKeyMgmt):
         return (cek, ek)
 
     def unwrap(self, key, ek):
-        rk = self.get_key(key, 'decrypt')
+        rk = self._get_key(key, 'decrypt')
 
         # Implement RFC 3394 Key Unwrap - 2.2.3
         # TODO: Use cryptography once issue #1733 is resolved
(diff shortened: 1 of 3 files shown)


npmccallum:
The second patch has a typo: algorythms.

jwcrypto/jwe.py
```diff
@@ -240,6 +240,55 @@ def unwrap(self, key, keylen, ek, headers):
...
+        tag = encryptor.tag
+        return {'cek': cek, 'ek': ek,
+                'header': {'iv': base64url_encode(iv),
+                           'tag': base64url_encode(tag)}}
+
+    def unwrap(self, key, keylen, ek, headers):
+        rk = self._get_key(key, 'decrypt')
+
+        if 'iv' not in headers:
+            raise InvalidJWEData('Invalid Header')
```
npmccallum: This exception (and the one below) seem awfully vague. Perhaps you could give more detailed descriptions?

jwcrypto/jwe.py
```diff
@@ -542,6 +600,25 @@ def _get_alg_enc_from_headers(self, jh):
         enc = self._jwa(encname)
         return alg, enc
 
+    def _encrypt(self, alg, enc, jh):
+        aad = base64url_encode(self.objects.get('protected', ''))
+        if 'aad' in self.objects:
+            aad += '.' + base64url_encode(self.objects['aad'])
```
npmccallum: Isn't aad already encoded? Unless I'm mistaken, I think you're double-encoding here.

jwcrypto/jwe.py
```diff
@@ -635,6 +701,20 @@ def serialize(self, compact=False):
                 rec = self.objects['recipients'][0]
             else:
                 rec = self.objects
+            if 'header' in rec:
```
npmccallum: This is ugly. Is there really not a better way to do this?

jwcrypto/jwe.py
```diff
@@ -289,6 +290,56 @@ def unwrap(self, key, keylen, ek, headers):
...
 
 
+class _Pbes2HsAesKw(_RawKeyMgmt):
+
+    def __init__(self, hashsize, keysize):
+        self.backend = default_backend()
+        self.hashsize = hashsize
+        self.keysize = keysize // 8
+
+    def _get_key(self, alg, key, p2s, p2c):
```
npmccallum: I don't like this. The input to PBES2 isn't an oct JWK it is a password string. It seems to me that conflating these two makes it easy to use a low entropy key where a salted, derived key should be used.

jwcrypto/jwe.py
```diff
@@ -485,6 +488,130 @@ def unwrap(self, key, keylen, ek, headers):
...
+        # AlgorithmID
+        otherinfo = struct.pack('>I', len(algID))
+        otherinfo += bytes(algID.encode('utf8'))
+
+        # PartyUInfo
+        if 'apu' in headers:
+            apu = base64url_decode(headers['apu'])
+        else:
+            apu = b''
+        otherinfo += struct.pack('>I', len(apu))
```
npmccallum: I think this is cleaner:
`apu = base64url_decode(headers['apu']) if 'apu' in headers else b''`

jwcrypto/jwe.py
```diff
@@ -485,6 +488,130 @@ def unwrap(self, key, keylen, ek, headers):
...
+        otherinfo += apu
+
+        # PartyVInfo
+        if 'apv' in headers:
```
npmccallum: Same as above.

jwcrypto/jwe.py
simo5: Will do

jwcrypto/jwe.py
simo5: Nope self.objects['add'] is in binary form throughout the code, it is not encoded on __init__ and it is decoded on deserialization.

jwcrypto/jwe.py
simo5: Seem the least invasive one, all other alternatives I could think of were more invasive and as ugly.

npmccallum:
Aside from my comments and your failing test, LGTM.

jwcrypto/jwe.py
simo5: I generally prefer the more verbose syntax, but I'll change this

jwcrypto/jwe.py
simo5: How so ?
Any key/password you pass in will be salted and derived.

npmccallum: `jwk = {"kty": "oct", "k": "thisismypassword"}`

If you accidentally pass that to, say, A128GCMKW, it will work; but be horribly insecure. I like the feature of PBES2 only accepting a string password and A128GCMKW only accepting a JWK. It provides type separation so that the key for one cannot be used in the other.

npmccallum: This is your library and ultimately you need to choose. But the above is my preference.

jwcrypto/jwe.py
npmccallum: I mean, can't you pawn this off on the caller? ISTM that placing iv/tag in the recipient header is the right move. This is what I do in jose...

jwcrypto/jwe.py
simo5: I would have to add a non-standard Key type to JWK to do that, because the JWE interface requires that key is a JWK object.
Alternatively I'd need to allow to pass in a string just for the PBES2 algorithm and forgo initial validation of the input ...

anyway see PR #40

PR #40, opened 2016-07-15
Allow both a key or a password for Pbes2

simo5:
@npmccallum check if this is good enough

### DIFF
--- a/jwcrypto/jwe.py
+++ b/jwcrypto/jwe.py
@@ -156,6 +156,8 @@ class _RSA(_RawKeyMgmt):
         self.padfn = padfn
 
     def _check_key(self, key):
+        if not isinstance(key, JWK):
+            raise ValueError('key is not a JWK object')
         if key.key_type != 'RSA':
             raise InvalidJWEKeyType('RSA', key.key_type)
 
@@ -215,6 +217,8 @@ class _AesKw(_RawKeyMgmt):
         self.keysize = keysize // 8
 
     def _get_key(self, key, op):
+        if not isinstance(key, JWK):
+            raise ValueError('key is not a JWK object')
         if key.key_type != 'oct':
             raise InvalidJWEKeyType('oct', key.key_type)
         rk = base64url_decode(key.get_op_key(op))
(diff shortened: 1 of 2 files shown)


simo5:
The reason why I still allow a key is that RFC 7518 says this:
   The PBES2 password input is an
   octet sequence; if the password to be used is represented as a text
   string rather than an octet sequence, the UTF-8 encoding of the text
   string MUST be used as the octet sequence.

I interpret this to mean random keys composed by non utf-8 sequences are also acceptable, and does can only be represented sanely with a "oct" key.

npmccallum:
Must you use a JWK for this? Python literally has a bytes type.

simo5:
I could use bytes/string vs unicode I guess, b ut it is easier to pass an octect key as a JWK, why invent yet another way to pass in a non printable key into this method ?

npmccallum:
Because I'm trying to create a hard conceptual barrier between cryptographically strong JWKs and cryptographically weak PBES2 inputs. The idea is to make it impossible for a consumer of the API to accidentally use a weak input to an algorithm like A128KW.

simo5:
So the issue here is that in python 2 bytes and string are the same thing, so given an input I can't distinguish between the two and can't validate.
Using a JWK when you want to pass in an octect string that is not a password is safer.

One option would be to ignore checks and rely on the user to pass in proper input.
In that case I can accept bytes or strings, in case of strings we'd get proper validation only if Python 3 is being used, in python 2 anything can slip in.

npmccallum:
You only need to validate on Python 3 so that you can perform encoding. On Python 2 you just expect people to pass in the proper thing.

simo5:
Take a look at this one, I just accept byes or string, and encode the string to bytes, the encoding happens only in python3 as python 2 always considers both strings and bytes as bytes

simo5:
Also if a jwk is passed in it fails with an error as JWK has no .encode() function

jwcrypto/jwe.py
```diff
@@ -390,9 +396,10 @@ def __init__(self, hashsize, keysize):
         self.keysize = keysize // 8
 
     def _get_key(self, alg, key, p2s, p2c):
-        if key.key_type != 'oct':
-            raise InvalidJWEKeyType('oct', key.key_type)
-        plain = base64url_decode(key.get_op_key('encrypt'))
+        if isinstance(key, bytes):
```
npmccallum: `plain = key if isinstance(key, bytes) else key.encode('utf8')`

npmccallum:
Besides my preferred style choice, LGTM.

jwcrypto/jwe.py
simo5: I thk in this case I really prefer the more verbose form, it's hard to understand what the compact form does at a glance

simo5:
Just changed the commit to make it more clear why we are breaking the Pbes2 interface.
Will push once tests mark all green

Issue #43, opened 2016-07-27
Utility functions for importing key files on disk?


wrouesnel:
A pretty standard use of this library is dealing with key files which need to be serialized/deserialized to disk. The JWK objects should either expose a constructor which can initialize them from on-disk PEM files, or else maybe some utility functionality should be provided to make this available?

A good middle-ground would be allowing directly initializing JWK's from rsa.Cryptography objects, since internally this is done anyway with functions like `JWK._rsa_pri`

simo5:
Isn't this just a dupliucate of #25 ?

simo5:
duplicate of #25

Issue #25, opened 2016-05-11
Load private / public key from PEM files


tiran:
PEM-encoded private key and X.509 cert files are common file formats for key material. jwcrypto should support loading private and public keys from these formats. With cryptography it is trivial, too.

I have a working implementation for Python 3, https://fedorapeople.org/cgit/cheimes/public_git/doregau.git/tree/authserver.py#n44

simo5:
Are you thinking about adding a way to import a PEM file content in a JWK object ?
A PEM file is more than just keys, but a certificate does not map 1:1 to a JWK object.

PR #45, opened 2016-08-03
Support Importing/Exporting PEM data

simo5:
With this functions PEM files can be easily imported, and JWK Keys can be easily exported into PEM files.
@tiran @wrouesnel please take a look at this one, it implements #25 / #43

### DIFF
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -387,23 +387,26 @@ class JWK(object):
 
     def export(self, private_key=True):
         """Exports the key in the standard JSON format.
+        Exports the key regardless of type, if private_key is False
+        and the key is_symmetric an exceptionis raised.
 
         :param private_key(bool): Whether to export the private key.
                                   Defaults to True.
         """
-        if private_key is not True:
+        if private_key is True:
+            # Use _export_all for backwards compatibility, as this
+            # function allows to export symmetrict keys too
+            return self._export_all()
+        else:
             return self.export_public()
-        d = dict()
-        d.update(self._params)
-        d.update(self._key)
-        d.update(self._unknown)
-        return json_encode(d)
 
     def export_public(self):
         """Exports the public key in the standard JSON format.
-           This function is deprecated and maintained only for
-           backwards compatibility, use export(private_key=False)
-           instead."""
+        It fails if one is not available like when this function
+        is called on a symmetric key.
+        """
+        if not self.has_public:
+            raise InvalidJWKType("No public key available")
         pub = {}
         preg = JWKParamsRegistry
         for name in preg:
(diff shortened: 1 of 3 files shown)


jwcrypto/jwk.py
```diff
@@ -549,12 +550,84 @@ def import_from_pyca(self, key):
         else:
             raise InvalidJWKValue('Unknown key object %r' % key)
 
+    def import_from_pem(self, data, password=None):
+        """Imports a key from data loaded from a PEM file.
```
tiran: I don't like the fact that the method attempts to load public and private keys. As an application developer I'd like to have more control over the key type and have dedicated methods to load either a private or a public key.

jwcrypto/jwk.py
```diff
@@ -549,12 +550,84 @@ def import_from_pyca(self, key):
...
+
+        :param private_key: Whether the private key should be exported.
+         Defaults to `False` which means the public key is exported by default.
+        :param password(bytes): A password for wrapping the private key.
+         Defaults to False which will cause the operation to fail. To avoid
+         encryption the user must explicitly pass None, otherwise the user
+         needs to provide a password in a bytes buffer.
+        """
+        e = serialization.Encoding.PEM
+        if private_key is True:
```
tiran: Please drop 'is True'

if private_key:

jwcrypto/jwk.py
```diff
@@ -549,12 +550,84 @@ def import_from_pyca(self, key):
...
+            return self._get_public_key().public_bytes(encoding=e, format=f)
+
     @classmethod
     def from_pyca(cls, key):
         obj = cls()
         obj.import_from_pyca(key)
         return obj
 
+    @classmethod
+    def from_pem(cls, data, password=None):
```
tiran: Is this the official API to create a JWK from PEM data? In that case ```import_from_pem()``` should be a private method.

simo5: I don't think a classmethod can acces private methods (last time I tried for from_pyca I got nasty warnings).
But in either case I am fine with both being public methods.

jwcrypto/jwk.py
simo5: Won't this always match anything that is not "False" or "None", I want to make sure there is a True there and no a bytes blob with a private key for example.

jwcrypto/jwk.py
simo5: So would you like a qualifier in the function signature?
Something like: import_from_pem(self, data, type=None|"private"|"public"|"x509", password=None) ?
Where None would be the default and autodetect what's in the file, but with the ability to force to check for a specific type ?

jwcrypto/jwk.py
```diff
@@ -405,6 +409,52 @@ def export_public(self):
...
+        It fails for a JWK that has only a public key or is symmetric.
+        """
+        if self.has_private:
+            return self._export_all()
+        raise InvalidJWKType("No private key available")
+
+    def export_symmetric(self):
+        if self.is_symmetric:
+            return self._export_all()
+        raise InvalidJWKType("Not a symmetric key")
```
tiran: I'd prefer a different order:
```
if not self.is_symmetric:
     raise InvalidJWKType("Not a symmetric key")
return self._export_all()
```

simo5: Why do you feel it is better this way ?

PR #51, opened 2016-08-19
Update docs and project URL

simo5:
Version is now at 0.3.0, docs were still sayion 0.1.0.
Also update setup.py to the correct URL for the project.


tiran:
ACK

PR #55, opened 2016-08-21
Allow to pass in dictionaries for headers

### LINKED ISSUE #53: JWE API feedback
First of all thanks for an excellent library. I'm using the library for JWE de/encryption in combination with square/js-jose.

It took me a while to figure out the API and what I ended up with is:
```
def encrypt(plaintext, public_jwk)
    protected = '{"alg":"RSA-OAEP","enc":"A256GCM"}'
    jwe = JWE(plaintext=plaintext, protected=protected)
    jwe.add_recipient(public_jwk)
    ciphertext = jwe.serialize(compact=True)
    return ciphertext

def decrypt(ciphertext, private_jwk):
    jwe = JWE()
    jwe.deserialize(ciphertext, key=private_jwk)
    plaintext = jwe.payload.decode('utf-8')
    return plaintext
```
Unfortunately those two functions look very different and aren't at all symmetric. If this is at all reasonable code, I found a couple of things non-obvious:

* The protected header could use an example in the docs. Otherwise you have to read large parts of the RFC to understand what it means.
* The protected argument to JWE is a string, I'd expected it to be a Python dict and assumed the API would take care of the conversion internally.
* It would be nice if there would be a key argument to ``JWE.__init__`` adding a default recipient. Since I use compact encoding, there can only be a single recipient and this would avoid the extra call to add_recipient.
* The JWE.deserialize function notes that it destroys the entire JWE state. Wouldn't it make more sense to change this into a class or staticmethod which internally creates a new JWE instance?
* JWE.deserialize has no return value. I expected it to return the plaintext and not get it from the payload property.

This is all well meant feedback after I used the library for a single purpose, so I might very well be missing larger implications or context.

Thanks! Hanno

simo5:
Where it makes sense, allow to pass in a dictionary for the various headers
and auto-encode them as needed.

Resolves #53

### DIFF
--- a/jwcrypto/jwe.py
+++ b/jwcrypto/jwe.py
@@ -864,10 +864,16 @@ class JWE(object):
         if aad:
             self.objects['aad'] = aad
         if protected:
-            json_decode(protected)  # check header encoding
+            if isinstance(protected, dict):
+                protected = json_encode(protected)
+            else:
+                json_decode(protected)  # check header encoding
             self.objects['protected'] = protected
         if unprotected:
-            json_decode(unprotected)  # check header encoding
+            if isinstance(unprotected, dict):
+                unprotected = json_encode(unprotected)
+            else:
+                json_decode(unprotected)  # check header encoding
             self.objects['unprotected'] = unprotected
         if algs:
             self.allowed_algs = algs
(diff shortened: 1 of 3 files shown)


simo5:
@hannosch please take a look

hannosch:
@simo5 Looks good to me.

PR #58, opened 2016-08-23
Move JWS/JWE crypto in its own JWA module

simo5:
This keeps all algorithms defined in RFC 7518 (JWA) in a separate module referenced by both JWS and JWE

@tiran please check this out


tiran:
LGTM

Issue #49, opened 2016-08-18
size type confusion


tiran:
jwcrypto sometimes uses bits and sometimes bytes for key size, e.g. JWK(generate='oct') takes size in bits, jwe._direct.wrap() reports InvalidCEKeyLength in bytes. Please stick to either bits or bytes.

PR #59, opened 2016-08-23
Allow to pass 'alg' as an argument for key generation

### LINKED ISSUE #50: JWK(generate='oct') should support algorithm argument
jwcrypto should support generating symmetric keys by algorithm name and auto-detect the required key size.

```
JWK(generate='oct', algorithm='A128CBC-HS256')
```

simo5:
This allows to automatically get a key of adequate size for the selected algorithm.

Resolves #50, depends on #58 

@tiran please check

### DIFF
--- a/jwcrypto/jwa.py
+++ b/jwcrypto/jwa.py
@@ -1105,12 +1105,16 @@ class JWA(object):
     }
 
     @classmethod
+    def instantiate_alg(cls, name, use=None):
+        alg = cls.algorithms_registry[name]
+        if use is not None and alg.algorithm_use != use:
+            raise KeyError
+        return alg()
+
+    @classmethod
     def signing_alg(cls, name):
         try:
-            obj = cls.algorithms_registry[name]()
-            if obj.algorithm_use != 'sig':
-                raise KeyError()
-            return obj
+            return cls.instantiate_alg(name, use='sig')
         except KeyError:
             raise InvalidJWAAlgorithm(
                 '%s is not a valid Signign algorithm name' % name)
@@ -1118,10 +1122,7 @@ class JWA(object):
     @classmethod
     def keymgmt_alg(cls, name):
         try:
-            obj = cls.algorithms_registry[name]()
-            if obj.algorithm_use != 'kex':
-                raise KeyError()
-            return obj
+            return cls.instantiate_alg(name, use='kex')
         except KeyError:
             raise InvalidJWAAlgorithm(
                 '%s is not a valid Key Management algorithm name' % name)
@@ -1129,10 +1130,7 @@ class JWA(object):
     @classmethod
     def encryption_alg(cls, name):
         try:
-            obj = cls.algorithms_registry[name]()
-            if obj.algorithm_use != 'enc':
-                raise KeyError()
-            return obj
+            return cls.instantiate_alg(name, use='enc')
         except KeyError:
             raise InvalidJWAAlgorithm(
                 '%s is not a valid Encryption algorithm name' % name)
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -227,11 +227,22 @@ class JWK(object):
 
         gen(params)
 
-    def _generate_oct(self, params):
-        size = 128
+    def _get_gen_size(self, params, default_size=None):
+        size = default_size
         if 'size' in params:
             size = params['size']
             del params['size']
+        elif 'alg' in params:
+            try:
+                from jwcrypto.jwa import JWA
+                alg = JWA.instantiate_alg(params['alg'])
+            except KeyError:
+                raise ValueError("Invalid 'alg' parameter")
+            size = alg.min_key_size
+        return size
+
+    def _generate_oct(self, params):
+        size = self._get_gen_size(params, 128)
         key = os.urandom(size // 8)
         params['kty'] = 'oct'
         params['k'] = base64url_encode(key)
@@ -243,13 +254,10 @@ class JWK(object):
 
     def _generate_RSA(self, params):
         pubexp = 65537
-        size = 2048
+        size = self._get_gen_size(params, 2048)
         if 'public_exponent' in params:
             pubexp = params['public_exponent']
             del params['public_exponent']
-        if 'size' in params:
-            size = params['size']
-            del params['size']
         key = rsa.generate_private_key(pubexp, size, default_backend())
         self._import_pyca_pri_rsa(key, **params)
 
--- a/jwcrypto/tests.py
+++ b/jwcrypto/tests.py
@@ -198,8 +198,12 @@ class TestJWK(unittest.TestCase):
         jwk.JWK.generate(kty='oct', size=256)
         jwk.JWK.generate(kty='RSA', size=4096)
         jwk.JWK.generate(kty='EC', curve='P-521')
-        k = jwk.JWK.generate(kty='oct', size=256, kid='MySymmetricKey')
+        k = jwk.JWK.generate(kty='oct', alg='A192KW', kid='MySymmetricKey')
         self.assertEqual(k.key_id, 'MySymmetricKey')
+        self.assertEqual(len(base64url_decode(k.get_op_key('encrypt'))), 24)
+        jwk.JWK.generate(kty='RSA', alg='RS256')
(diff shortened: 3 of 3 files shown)


jwcrypto/jwa.py
```diff
@@ -44,6 +56,22 @@ def algorithm_use(self):
         raise NotImplementedError
 
 
+# Note: l is the number of bits, which should be a multiple of 16
+def _encode_int(n, l):
+    e = hex(n).rstrip("L").lstrip("0x")
```
tiran: '{:x}'.format(n)

jwcrypto/jwa.py
```diff
@@ -44,6 +56,22 @@ def algorithm_use(self):
         raise NotImplementedError
 
 
+# Note: l is the number of bits, which should be a multiple of 16
+def _encode_int(n, l):
```
tiran: l is bad name for a var because it looks very similar to 1: l + 7 and 1 + 7

jwcrypto/jwa.py
```diff
@@ -44,6 +56,22 @@ def algorithm_use(self):
...
+    e = hex(n).rstrip("L").lstrip("0x")
+    elen = len(e)
+    ilen = ((l + 7) // 8) * 2  # number of bytes rounded up times 2 chars/bytes
+    if elen > ilen:
+        e = e[:ilen]
+    else:
+        e = '0' * (ilen - elen) + e  # pad as necessary
```
tiran: The whole block can be replaced by e.rjust(ilen, '0')[:ilen]

jwcrypto/jwa.py
```diff
@@ -466,6 +488,991 @@ def algorithm_use(self):
...
+        if len(cek) != keylen:
+            raise InvalidJWEKeyLength(keylen, len(cek))
+        return cek
+
+
+class _Rsa15(_RSA, JWAAlgorithm):
+    def __init__(self):
+        super(_Rsa15, self).__init__(padding.PKCS1v15())
+
+    @property
```
tiran: Why are you using properties instead of class attributes here? All values are static. If you put them in class vars, then the values can be introspected on the class, too.

jwcrypto/jwa.py
```diff
@@ -485,15 +1492,61 @@ class JWA(object):
...
+        'A256GCM': _A256Gcm
     }
 
     @classmethod
     def signing_alg(cls, name):
         try:
             obj = cls.algorithms_registry[name]()
             if obj.algorithm_use != 'sig':
-                raise InvalidJWAAlgorithm(name)
+                raise KeyError()
```
tiran: raise KeyError(name)

ValueError or TypeError might be better. After all the algorithm is found but it has wrong type.

jwcrypto/jwa.py
```diff
@@ -485,15 +1492,61 @@ class JWA(object):
...
+        except KeyError:
+            raise InvalidJWAAlgorithm(
+                '%s is not a valid Signign algorithm name' % name)
+
+    @classmethod
+    def keymgmt_alg(cls, name):
+        try:
+            obj = cls.algorithms_registry[name]()
+            if obj.algorithm_use != 'kex':
+                raise KeyError()
```
tiran: dito

jwcrypto/common.py
```diff
@@ -46,3 +46,57 @@ def __init__(self, message=None):
...
+
+class InvalidCEKeyLength(Exception):
+    """Invalid CEK Key Length.
+
+    This exception is raised when a Content Encryption Key does not match
+    the required lenght.
+    """
+
+    def __init__(self, expected, obtained):
+        msg = 'Expected key of length %d, got %d' % (expected, obtained)
```
tiran: Is this length in bites or bytes? See #49

jwcrypto/common.py
```diff
@@ -46,3 +46,57 @@ def __init__(self, message=None):
...
+
+class InvalidJWEKeyLength(Exception):
+    """Invalid JWE Key Length.
+
+    This exception is raised when the provided JWK Key does not match
+    the lenght required by the sepcified algorithm.
+    """
+
+    def __init__(self, expected, obtained):
+        msg = 'Expected key of lenght %d, got %d' % (expected, obtained)
```
tiran: bits or bytes?

jwcrypto/jwa.py
simo5: This code is just moved, please open an issue for problems with the original code

jwcrypto/jwa.py
simo5: See above

jwcrypto/jwa.py
simo5: See above

jwcrypto/jwa.py
simo5: I guess we could do that, but given not all properties are static, I wanted to keep them all properties so they are consistent thoughout.
Can you override a class attribute with a property in a subclass ?

jwcrypto/jwa.py
simo5: Well the idea is that wehn you pass in a use you restrict the virtual set of algorithms used to the subset markled sig, so it's like the other algorithms do not exist for the caller. So I think KeyError is appropriate here (think of it as a virtual sub-dict is being queried). Any error we throw is not seen directly bu the caller anyway as we catch it next and wrap it in a InvalidJWAAlgorithm exception with the proper name and we do not expose the original exception though.

jwcrypto/common.py
simo5: It's all; supposed to be bits, I'll change this error message too to make it clear

simo5: Correction, that is done in PR #60, which depends on this one.

jwcrypto/common.py
simo5: See #60

simo5:
For posterity, all these comments (and replies) really belong to PR #58

jwcrypto/jwa.py
```diff
@@ -1105,34 +1105,32 @@ class JWA(object):
     }
 
     @classmethod
+    def instantiate_alg(cls, name, use=None):
+        obj = cls.algorithms_registry[name]()
+        if use and obj.algorithm_use != use:
```
tiran: please check with ```if use is not None and ...```

jwcrypto/jwk.py
```diff
@@ -243,13 +254,10 @@ def _encode_int(self, i):
 
     def _generate_RSA(self, params):
         pubexp = 65537
-        size = 2048
+        size = self._get_gen_size(params, 2048)
         if 'public_exponent' in params:
             pubexp = params['public_exponent']
             del params['public_exponent']
```
tiran: pubexp = params.pop('public_exponent') gets and deletes the entry from the dict at the same time

jwcrypto/jwk.py
```diff
@@ -227,11 +227,22 @@ def generate_key(self, **params):
 
         gen(params)
 
-    def _generate_oct(self, params):
-        size = 128
+    def _get_gen_size(self, params, default_size=None):
+        size = default_size
         if 'size' in params:
             size = params['size']
             del params['size']
```
tiran: size = params.pop('size')

jwcrypto/jwk.py
simo5: This is an antypattern found elsewhere in the jwk code.
I'll fix it all at once once this PR is pushed

jwcrypto/jwk.py
simo5: as above tracked in #62

jwcrypto/jwa.py
```diff
@@ -1105,34 +1105,32 @@ class JWA(object):
     }
 
     @classmethod
+    def instantiate_alg(cls, name, use=None):
+        obj = cls.algorithms_registry[name]()
+        if use is not None and obj.algorithm_use != use:
+            raise KeyError
+        return obj
```
tiran: With the class attributes you could delay instantiating:

```
impl = cls.algorithms_registry[name]
if use is not None and impl.algorithm_use != use:
    raise KeyError
return impl()
```

simo5: yeah, I think I had it this way before and then canged it back nd forth a few times, let me change it

tiran:
ACK

PR #60, opened 2016-08-24
Always use bit sizes for keys through the code

### LINKED ISSUE #49: size type confusion

simo5:
Makes all the JWA code and error messages use bit sizes for key material.

Resolves #49, depends on #58

@tiran please check

### DIFF
--- a/jwcrypto/jwa.py
+++ b/jwcrypto/jwa.py
@@ -44,8 +44,8 @@ class JWAAlgorithm(object):
         pass
 
     @abc.abstractproperty
-    def min_key_size(self):
-        """The recommended/minimum key size or bit strenght"""
+    def keysize(self):
+        """The actual/recommended/minimum key size"""
         pass
 
     @abc.abstractproperty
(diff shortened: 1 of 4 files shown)


jwcrypto/jwa.py
```diff
@@ -599,10 +597,7 @@ class _Pbes2HsAesKw(_RawKeyMgmt):
 
     def __init__(self):
         self.backend = default_backend()
-
-    @property
-    def min_key_size(self):
-        return self.keysize * 8
+        self.aeskwmap = {128: _A128KW, 192: _A192KW, 256: _A256KW}
```
tiran: Since this class is defined after the AES classes, you can move the dict into the class body. I'd prefix it was _ to make clear that it is an internal API.

tiran:
ACK, you could move the aeskwmap or wait until we find a better way. IDC

simo5:
I'll leave changes to aeskwmap to a later time, I also thought hard about whther to make it a lcass attribute or something else. no option stood up as better than any other so I'll just defer a decision until we have something that makes one of the options better.

Issue #65, opened 2016-08-31
CVE-2016-6298: Million Messages Attack vulnerability


simo5:
The jwcrypto implementation of the RSA1_5 algorithm is vulnerable to the Million Message Attack described in [RFC 3128](https://www.ietf.org/rfc/rfc3218.txt).

A timing attack can be leveraged against the implementation to detect when a chosed ciphertext generates a valid header and padding because invalid headr/padding generates a code exception and cryptographic operations are terminated earlier resulting in measurably faster processing over the network.

Many thanks to Dennis Detering <dennis.detering@rub.de> for discovering and reporting this vulnerability.

PR #66, opened 2016-08-31
Fix for CVE-2016-6298

simo5:
This fixes the security issue reported in #65

### DIFF
--- a/jwcrypto/jwa.py
+++ b/jwcrypto/jwa.py
@@ -379,6 +379,23 @@ class _Rsa15(_RSA, JWAAlgorithm):
     def __init__(self):
         super(_Rsa15, self).__init__(padding.PKCS1v15())
 
+    def unwrap(self, key, bitsize, ek, headers):
+        self._check_key(key)
+        # Address MMA attack by implementing RFC 3218 - 2.3.2. Random Filling
+        # provides a random cek that will cause the decryption engine to
+        # run to the end, but will fail decryption later.
+
+        # always generate a random cek so we spend roughly the
+        # same time as in the exception side of the branch
+        cek = _randombits(bitsize)
+        try:
+            cek = super(_Rsa15, self).unwrap(key, bitsize, ek, headers)
+            # always raise so we always run through the exception handling
+            # code in all cases
+            raise Exception('Dummy')
+        except Exception:  # pylint: disable=broad-except
+            return cek
+
 
 class _RsaOaep(_RSA, JWAAlgorithm):
 
--- a/jwcrypto/tests.py
+++ b/jwcrypto/tests.py
@@ -842,6 +842,86 @@ class TestJWE(unittest.TestCase):
             e = jwe.JWE(algs=['A256KW'])
             e.deserialize(E_A5_ex, E_A4_ex['key2'])
 
+MMA_vector_key = jwk.JWK(**E_A2_key)
+MMA_vector_ok_cek =  \
+    '{"protected":"eyJlbmMiOiJBMTI4Q0JDLUhTMjU2In0",' \
+    '"unprotected":{"jku":"https://server.example.com/keys.jwks"},' \
+    '"recipients":[' \
+    '{"header":{"alg":"RSA1_5","kid":"2011-04-29"},' \
+    '"encrypted_key":'\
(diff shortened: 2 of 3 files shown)


jwcrypto/tests.py
```diff
@@ -842,6 +842,85 @@ def test_A5(self):
...
+            ok_e = jwe.JWE()
+            ok_e.deserialize(MMA_vector_ok_cek)
+            ko_cek = 0
+            ko_e = jwe.JWE()
+            ko_e.deserialize(MMA_vector_ko_cek)
+
+            import time
+
+            for _ in range(self.iterations):
+                start = time.clock()
```
tiran: time.clock() is the worst clock on Unix. You want:

* time.perf_counter() on Python 3.3+
* time.time() on POSIX
* time.clock() on Windows

jwcrypto/jwa.py
```diff
@@ -379,6 +379,23 @@ class _Rsa15(_RSA, JWAAlgorithm):
...
+        self._check_key(key)
+        # Address MMA attack by implementing RFC 3218 - 2.3.2. Random Filling
+        # provides a random cek that will cause the decryption engine to
+        # run to the end, but will fail decryption later.
+
+        # always generate a random cek so we spend roughly the
+        # same time as in the exception side of the branch
+        cek = _randombits(bitsize)
+        try:
+            cek = super(_Rsa15, self).unwrap(key, bitsize, ek, headers)
```
tiran: Is _RSA.unwrap() and private key decryption constant time? Please add comments to all critical paths.

jwcrypto/jwa.py
```diff
@@ -379,6 +379,23 @@ class _Rsa15(_RSA, JWAAlgorithm):
...
+            # always raise so we always run through the exception handling
+            # code in all cases
+            raise Exception('Dummy')
```
tiran: exception chaining and depths of the stack frame may cause different timings. It's fragile.

tiran:
I'm not fully convinced that the mitigation removes all timing variations. It's almost impossible to write constant timing code in Python. Heck, it's even extremely hard in C.

How about we mark PKCS1v15, NULL and other bad algos as insecure and not offer them by default? I have an idea to make the registry more flexible.

jwcrypto/tests.py
simo5: This is just a tool to get a rough idea of the timing, the variation is still high enough to not really matter I think.

jwcrypto/jwa.py
simo5: The openssl implementation of the RSA decrypt function uses contant time code for the critical parts, but it is not constant time in general, and doesn't need to be.

jwcrypto/jwa.py
simo5: It does, in my testing the difference between valid CEK and invalid CEKis still measurable in a tight lopp on the local machine, but hardly over the network where the rest of the stack noise is greater than the timing difference. raising an exception there lowered the difference between the code path of about .4% on average in my testing, so I added it. It is not perfect, but can't be,

Note that I measured up to 3% variability in decoding 2 identical messages encrypted with difference CEKs, so as long as we stay in the error margin the attacker has too much noise to figure out what he got.

simo5:
We can't remove all timing variations but that is not the point, we just need to avoid big enough variations to be measurable over the network. Different messages already cause different timings on their own.
It makes no sense to remove RSA1_5 because we have a mitigation and the vulnerability is usage specific, it is only present if an attacker can capture messages and use a server as an oracle. There are perfectly valid uses of this algorithm and it is mandated by the standard,

Merenon:
This patch looks pretty good to me. It's a great idea to always throw an exception: My test cases showed great results with this fix and I believe the difference in timing is negligible over the network.

I agree, that RSA 1.5 is not the best algorithm to support, but as Simo already mentioned, it's necessary to support it in order to be compliant with the specification.
I guess, there are some discussions about deprecating this algorithm though...

Thanks for your fast reaction in fixing this, Simo! It was a pleasure!

rokclimb15:
@simo5 I'm working on backporting the fix to Ubuntu 16.04 (0.2.1). Could you possibly share your tests for the issue so I can write a unit test and confirm the backported fix works?

https://people.canonical.com/~ubuntu-security/cve/2016/CVE-2016-6298.html

PR #90, opened 2017-10-09
Fix error interpolation on nonstandard jwt claims

michaelboulton:
If the claim passed was not one of sub, aud, iat, etc. then passing a string as a value to check_claims would raise an error if it was not an integer


simo5:
Looks good, can you squash the commits ?

michaelboulton:
Done

PR #94, opened 2017-11-13
Make all exception derive from JWException

### LINKED ISSUE #93: Making all the exceptions inherit from JWException
It seems like this is just something that was missed out by accident? It would be nice if all the exceptions for JWE/JWT/etc. inherited from `JWException` in common.py so they could be caught in one except block.

simo5:
Fixes #93

### DIFF
--- a/jwcrypto/common.py
+++ b/jwcrypto/common.py
@@ -16,12 +16,12 @@ def base64url_encode(payload):
 
 
 def base64url_decode(payload):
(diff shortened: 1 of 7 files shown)


simo5:
@michaelboulton @tiran PTAL

tiran:
CI is using a new version of flake8 which is complaining about additional problems.

Issue #96, opened 2017-11-15
No tests for header or claims as string


ajmaddaford:
When creating a JWT `header` and `claims` can be passed as a string, but there don't seem to be any tests that test this. 

Also, when `claims` is passed as a string, any `default_claims` are not used (https://github.com/latchset/jwcrypto/blob/master/jwcrypto/jwt.py#L228). Should an exception be raised when rather than silently ignoring the default_claims?

simo5:
The default_claims are not added intentionally when claims are passed in as a string, why do you think an exception should be raised ?

simo5:
I guess that what we could do is to json_decode the claims instead, and then add default_claims perhaps ?
Is this what you expected to happen ?

ajmaddaford:
Thanks for the quick response! 

I would expect either it should use the provided default claims, or let me know that it isn't doing what I think it will e.g. by raising an exception.

It's not needed for how I'm using it, just something I noticed when changing from using a dict to a string for the claims (our dict can have Decimal objects in it so we use simplejson to convert it to a string).

simo5:
OK, thank you for the report, yeah I think that since we added the default_claims we should really unfold provided claims as a string and try to add the default ones, and raise if somethign fails.
I'll prepare a patch

ad-m:
Closeable?

simo5:
Will autoclose when #103 is merged

PR #98, opened 2017-11-15
Always augment claims if defaults are provided

simo5:


simo5:
See #96

simo5:
@tiran PTAL

simo5:
@tiran, added test, PTAL

PR #100, opened 2017-11-30
Add example of encrypting/decrypting using asymmetric keys

michaelboulton:
Just add a quick example on how to use asymmetric keys with JWE tokens.


simo5:
Is it useful to use a public_key and a private_key in this example ?
Should we just use a "key" ?

tiran:
We should enable doc tests to verify that our documentation examples don't get broken.

PR #101, opened 2017-11-30
Add function to get a public only JWK

### LINKED ISSUE #99: Extracting a public JWK from a private one
Assuming I generate a private key:
```python
private_key = JWK.generate(kty='RSA', size=4096)
```
It seems like the quickest way to get just a public version at the moment is to do this:
```python
public_key = JWK()
public_key.import_key(**json_decode(private_key.export_public()))
```
It would be good if there was a method called something like `extract_public` that had this behaviour:
```python
public_key = private_key.extract_public()
```
To make this simpler

simo5:
This makes it simpler to get a JWK that contains excluively a pulic
key from an existing JWK. It fails if no public key is available in
the source JWK.
Only known public elements are returned. "Unknown" attributes are
not copied over.

Resolves #99

### DIFF
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -405,6 +405,10 @@ class JWK(object):
         It fails if one is not available like when this function
         is called on a symmetric key.
         """
+        pub = self._public_params()
+        return json_encode(pub)
+
+    def _public_params(self):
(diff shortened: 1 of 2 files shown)


simo5:
@tiran, added tests, PTAL

PR #103, opened 2017-12-05
Test header using string not dict

### LINKED ISSUE #96: No tests for header or claims as string

simo5:
Change one test to also test using strings instead of dicts as headers.
This is supported in the code but was not tested.

Closes #96


simo5:
@tiran, oneliner to extend coverage, PTAL

simo5:
@tiran PTAL and merge if ok

PR #111, opened 2018-03-17
Fixing "A128KW" under "_EcdhEsAes256Kw"

enugentdt:
There was a typo in the "_EcdhEsAes256Kw" class. I have changed "A128KW" in the description for the class to become "A256KW" in accordance with the algorithm.


simo5:
Nice catch

PR #114, opened 2018-04-02
Rfc6979 to dss

### LINKED ISSUE #113: CryptographyDeprecationWarning in jwa.py
When using JWCrypto methods on my Python 3.6.4 installation, the process raises CryptographyDeprecationWarnings due to deprecated methods :

/usr/local/lib/python3.6/dist-packages/jwcrypto/jwa.py:153: CryptographyDeprecationWarning: encode_rfc6979_signature is deprecated and will be removed in a future version, use encode_dss_signature instead.
  int(hexlify(r), 16), int(hexlify(s), 16))

/usr/local/lib/python3.6/dist-packages/jwcrypto/jwa.py:144: CryptographyDeprecationWarning: decode_rfc6979_signature is deprecated and will be removed in a future version, use decode_dss_signature instead.
  r, s = ec_utils.decode_rfc6979_signature(signature)

For future releases, it would be interesting to fix this so the library will work with future Python releases.

Btw, your library is very awesome :)

simo5:
Fixes #113

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,13 +21,14 @@ testlong: export JWCRYPTO_TESTS_ENABLE_MMA=True
 testlong: export TOX_TESTENV_PASSENV=JWCRYPTO_TESTS_ENABLE_MMA
 testlong:
 	rm -f .coverage
-	tox -e py35
+	tox -e py36
 
 test:
 	rm -f .coverage
 	tox -e py27
 	tox -e py34 --skip-missing-interpreter
 	tox -e py35 --skip-missing-interpreter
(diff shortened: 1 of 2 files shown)


simo5:
@tiran PTAL

tiran:
LGTM, but you may have to bump minimal requirements or add some fallback for old PyCA cryptography.

simo5:
@tiran I checked I think the "new" method is available in Pyca all the way back to the last supported version.
I think it is ok to unsupport anything older ?

PR #115, opened 2018-04-10
Add imports for modules in package

jdemilledt:
This allows for someone to `import jwcrypto` and then use all of the modules in it without reimporting.

### DIFF
--- a/jwcrypto/__init__.py
+++ b/jwcrypto/__init__.py
@@ -1,0 +1,1 @@
+from . import common, jwa, jwe, jwk, jws, jwt  # NOQA: F401


simo5:
looks good, please squash together the commits and provide a commit message that explains the change

simo5:
Ah I see the commit mesage in the first commit is adequate, just squash them all

jdemilledt:
I do not have the option to squash.

jdemilledt:
Nevermind, I see it's something I have to do with the CLI.

jdemilledt:
Squashed.

simo5:
@tiran ok to merge this one ?

tiran:
-1

Except for rare, documented cases like ``os.path``, it is common practice to import submodules explicitly. The approach is also against the Zen of Python (name spaces, explicit is better than implicit). More importantly it slows down command line scripts that use only parts of jwcrypto.

jdemilledt:
@tiran This is not an issue in this case as importing a submodule of `jwcrypto` does not trigger the main package's `__init__.py`.

jdemilledt:
Disregard. Should have verified that first.

simo5:
@jdemilledt given the comment from @tiran do you still want to pursue this PR ?
I also thought a direct module import for jwe or jwa (etc..) would not cause this code to be called ...

simo5:
Sorry @jdemilledt after some off band discussion with @tiran I decided the tradeoff is not really worth it.
Thanks for the PR anyway, it was a good try :-)

jdemilledt:
I still wish to pursue this as code readability is increased when the exact package is referenced, thus removing the need to track down imports in large files.

simo5:
Ok if you can find a way to do this w/o impacting clients that wish to import only a specific subpackage we can try to do it.
Maybe an import jwcrypto.all trick of sorts ?

jdemilledt:
One thing I'd like to mention is that for importing a subpackage of JWCrypto, the difference is less than a tenth of a second. Is this acceptable?

jdemilledt:
It seems that the fundamental purpose of having the imports at the root is destroyed with any other methods.

jdemilledt:
@simo5 It seems my squashing of my commits to provide a better solution caused the PR to drop. Mind checking out my patch-1 branch and giving feedback?

jdemilledt:
After some testing, I have concluded that that hacky solution is no faster than just importing at the base. Many other packages do this, and everybody likes them. It'd be safer to just import at the base.

simo5:
On Thu, 2018-04-12 at 21:48 -0700, Julian DeMille wrote:
> It'd be safer to just import at the base.

Unfortunately GitHub is quite dumb when it comes to some invasive rebases and
the PR is clsoed, so it lost the thread. You'll need to open a new PR.
However I want to defer to @tiran on whether this is ok to do, he seemed pretty
against it as a general practice.

Simo.

tiran:
I think that implicit imports of submodules isn't a good idea, because it's against the Zen of Python. Barry thinks so, too. https://github.com/python/peps/issues/620#issuecomment-381237943

Barry and me have a total of 40 years experience with Python. Please trust our gut feeling.

jdemilledt:
I figured as much. It's also not as safe.

Issue #116, opened 2018-05-10
no simple way to import symmetric key


hmoffatt:
jwk.JWK has an export_key() method which returns a string containing the JSON, but there doesn't seem to be a way to import that format back.

import_key() seems to expect a dict. I got this to work with

k2 = import_key(**json.loads(key.export()))

simo5:
This is how the interface currently is expected to be used.
I guess we culd add a from_json() like we have on JWKSet ? Would that work ?

hmoffatt:
That works. 

JWKSet has the object method import_keyset() so it would be good if JWK also had an import_key which took a JSON string, but at least it's possible with this change.

simo5:
The JWK already has an import_key method, which expects an array of arguments, I cannot change that as it would break backwards compatibility.
I would theorethically check if a user is passing in just one argument and then try to see if it is a json string, but that is magic behavior.

JWKSet already exposes a class method named from_json(), so adding it to JWK is fine.
A class method is also better because JWK object resuse is discouraged.

hmoffatt:
All good

PR #118, opened 2018-05-10
Add method to import a JWK directly from json

simo5:
JWK.from_json(<json_string>) will import a previously exported key.

Fixes: #116


simo5:
@hmoffatt take a look
@tiran PTAL

jwcrypto/jwk.py
```diff
@@ -387,6 +387,19 @@ def import_key(self, **kwargs):
...
+    def from_json(cls, key):
+        """Creates a RFC 7517 JWK from the standard JSON format.
+
+        :param key: The RFC 7517 representation of a JWK.
+        """
+        obj = cls()
+        try:
+            jkey = json_decode(key)
+        except Exception:  # pylint: disable=broad-except
+            raise InvalidJWKValue()
```
tiran: Python 2 doesn't contain the exception context. Could you please include the inner exception inside the outer exception (``except Exception as e: raise InvalidJWKValue(e)``.

simo5: Good catch

PR #119, opened 2018-05-11
Make the default iterator return keys directly

### LINKED ISSUE #117: no way to get keys from JWKSet with no kid
You can add keys to a JWKSet which have no kid, but there is no way to get them out again.

simo5:
It has no value to return the 'keys' attribute itself, instead we now
we return the inner 'keys' object iterator when the JWKSet is iterated
over.
This way iterating over the JWKSet object returns its JWK objects.

Fixes #117


simo5:
@hmoffatt take a look
@tiran PTAL

hmoffatt:
Thanks looks good to me

PR #121, opened 2018-05-26
Allow bytestrings to be used to decode JWE tokens

### LINKED ISSUE #120: Attempting to use a bytestring as a secret breaks JWK
I'm attempting to decrypt a JWE token generated by someone else using the Ruby JOSE library, and they chose to encrypt it not with their secret, but with the SHA256 digest of their secret. 

In Ruby, this works fine, and can be used to generate a payload.

In Python, with jwcrypto, taking the secret and generating a SHA256 digest generates a bytestring. Handing a bytestring to JWK does not error, but handing that JWK object to JWE.decrypt() blows up with a UnicodeDecodeError.

Pull request with a test showing the issue and a bad solution to discuss shortly.

freyley:
Resolves #120 

I don't think this is the right solution, or the right place to put the tests. Advice on a better solution or rework on this PR welcome.

### DIFF
--- a/jwcrypto/common.py
+++ b/jwcrypto/common.py
@@ -23,7 +23,12 @@ def base64url_decode(payload):
         payload += '='
     elif size != 0:
         raise ValueError('Invalid base64 string')
-    return urlsafe_b64decode(payload.encode('utf-8'))
+    try:
+        return urlsafe_b64decode(payload.encode('utf'))
+    except (UnicodeDecodeError):
+        return payload
+
+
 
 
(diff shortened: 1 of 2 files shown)


simo5:
The patch is not ok, the test should pass any bytestring and generating the JWK should fail right then with an InvalidJWKValue.
The import_key() function should check that the passed key is an actual base64url encoded value I guess, and fail if not.
I'll open a PR to add that kind of validation.

simo5:
I think #122 properly handles this

PR #122, opened 2018-05-26
Validate key parameters on import

### LINKED ISSUE #120: Attempting to use a bytestring as a secret breaks JWK

simo5:
Fixes #120

### DIFF
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -25,21 +25,28 @@ JWKTypesRegistry = {'EC': 'Elliptic Curve',
 # RFC 7518 - 7.5
 # It is part of the JWK Parameters Registry, but we want a more
 # specific map for internal usage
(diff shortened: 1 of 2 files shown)


freyley:
@simo5 You can close #120, but you have not handled the broader problem of library interoperability. Please see my latest comment.

simo5:
@freyley I think my comment explains how to properly interoperate.

PR #123, opened 2018-06-25
Rework internal JWK structures for clarity

simo5:
This uses more idiomatic types to handle internal structured data


simo5:
@tiran PTAL

PR #126, opened 2018-08-06
Add support for RFC7797

### LINKED ISSUE #4: Add option to pass payload as a separate argument when checking signatures
The JWS spec apparently allow applications to send the signature (in all forms: compact, general, flattened) without the payload, and let the application provide the payload for verification.
JWS.deserialize() currently expects the payload to be always in the token.

simo5:
Together with non encoded payloads this commit also adds sull support
for dealing with detached payloads on deserlization and serialization.

The payload must still be provided for any computation, but can be
removed before serialization and add after deserialization of a JWS with
detached payload.

Fixes #4

### DIFF
--- a/jwcrypto/jws.py
+++ b/jwcrypto/jws.py
@@ -1,5 +1,7 @@
 # Copyright (C) 2015 JWCrypto Project Contributors - see LICENSE file
 
+from collections import namedtuple
+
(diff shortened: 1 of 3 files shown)


simo5:
@tiran PTAL

simo5:
Ok merging this as is, been waiting too long

PR #128, opened 2018-09-06
Properly store allowed algorithms when provided as argument

JonathanHuot:


simo5:
Nice catch!
Could you please also add a test that maes sure the alg flag is properly honored ?

Like setting only one specific alg and then trying to use it as well as trying to use one that should be denied and making sure you get the right behavior in both cases.

Issue #130, opened 2018-09-12
JWK.from_json returns None


adnanshussain:
I think it's because you are doing 

`return obj.import_key(**jkey)`

instead of 

`return obj`

simo5:
thanks, we should do that indeed ...

simo5:
Fixed in #132

PR #131, opened 2018-09-13
Fix JWK.from_json #130

JonathanHuot:
Instead of changing `JWK.from_json` from 
```
return obj.importXX
```
 to 
```
obj.import
return obj
```

I found better to change the signature of the `importXX` functions to `return self`.

### DIFF
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -429,6 +429,7 @@ class JWK(object):
                         raise InvalidJWKValue('Incompatible "use" and'
                                               ' "key_ops" values specified at'
                                               ' the same time')
+        return self
 
     @classmethod
     def from_json(cls, key):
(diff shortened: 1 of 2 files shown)


simo5:
Sorry @JonathanHuot but I think the correct fix here is to just return obj directly from from_json() and not so change the signature of import functions.

If you want to ament your commit to do that I'll glady merge.

simo5:
If you amend the commit please add a better subject and then add the string 'Fixes #130' on a single line in the commit comment

JonathanHuot:
Sorry, I forgot to tell the rationale behind this. Both this PR and #130 suggestion will fix `from_json()`.

However, if we implement the current PR, we add a `return` value of functions which didn't have one. So we don't break any implementations.

If we do the suggestion in #130, it means that, as a rule, all `import_*` functions should not have a `return` value. So the rest of the code should be in adequation, like the existing `JWKSet.import_keyset()` where it currently returns `self`. If any implementations use `import_keyset()` it will break them. 

That being said, `return self` is a bit too js-ish to me, so I have no particular concerns of removing those.

simo5:
I do not want to break people using import_keyset() unfortunately that accident happened historically.
We just need to fix from_json()
We might raise add a deprecation warning on using the return from import_keyset() and down the road remove it, but not right away.

simo5:
Close in favor of #132

Issue #136, opened 2018-10-26
ECDH-ES+A128KW doesn't work when CEK size is not 128 bits


Feaisil:
Hi there,

I have failures when I am trying to encrypt or decrypt a JWE with key wrapped ECDH-ES when the key wrapping key size is different from the content encryption key size, for instance with ECDH-ES+128KW and A128CBC-HS256 (CEK size 256):
```python
# encrypt using ECDH-ES+A128KW and A128CBC-HS256
key=jwk.JWK.generate(kty='EC', crv='P-256')
payload = "My Encrypted message"
jwetoken = jwe.JWE(payload.encode('utf-8'),
                       json_encode({"alg": "ECDH-ES+A128KW",
                                    "enc": "A128CBC-HS256"}))
jwetoken.add_recipient(key)
enc = jwetoken.serialize()
```

This leads to the following error:
```
File "<stdin>", line 1, in <module>
  File "[...]/jwcrypto/jwe.py", line 239, in add_recipient
    self._encrypt(alg, enc, jh)
  File "[...]/jwcrypto/jwe.py", line 195, in _encrypt
    iv, ciphertext, tag = enc.encrypt(self.cek, aad, data)
  File "[...]/jwcrypto/jwa.py", line 882, in encrypt
    cipher = Cipher(algorithms.AES(ekey), modes.CBC(iv),
  File "[...]/cryptography/hazmat/primitives/ciphers/algorithms.py", line 35, in __init__
    self.key = _verify_key_size(self, key)
  File "[...]/cryptography/hazmat/primitives/ciphers/algorithms.py", line 21, in _verify_key_size
    len(key) * 8, algorithm.name
ValueError: Invalid key size (0) for AES.
```
likewise, when decrypting. 

It looks like the key agreement algorithm overrides the content encryption key size with the key wrapping key size when calling the key wrapping/unwrapping.

There's no issue with the key wrapping or key agreement algorithms when not combined (ECDH-ES and A\*KW are working perfectly fine with any of the A\*CBC-HS\* algorithms)

cheers

Feaisil:
On decryption, I believe the issue is:
https://github.com/latchset/jwcrypto/blob/d0c3ee2797d1321304a423a46dac620e238f98f3/jwcrypto/jwa.py#L794
```
            bitsize = self.keysize
[...]
        dk = self._derive(key.get_op_key('unwrapKey'),
                          epk.get_op_key('wrapKey'),
                          alg, bitsize, headers)
[...]

PR #138, opened 2018-10-30
Fix case where CEK is greater in size than KEK in ECDH-ES

simo5:
See #136


simo5:
Added your JWE/JWK as a contributed test too,
thank you!

PR #144, opened 2018-12-11
Allow to use JWKSet on a JWT with no KID

### LINKED ISSUE #143: Cannot validate JWT with no kid in its header
Hi folks! First of all, I appreciate your library a lot. 

Recently, I ran into the following issue. A 3rd party system provides us a JWT which we want to verify. We have no means of changing implementation on the provider side. 
They send us a token which does not contain a _kid_ in its header. According to my understanding of RFC 7515, 7517 that field is optional. 

However, I got an exception on `jwt = JWT(jwt=token, key=jwks)`

```
  File "<...>/python3.6/site-packages/jwcrypto/jwt.py", line 474, in deserialize
    raise JWTMissingKeyID('No key ID in JWT header')

jwcrypto.jwt.JWTMissingKeyID: No key ID in JWT header
```

As I checked the code it's quite obvious that it does not work without "kid" in a header:
`./eggs/jwcrypto-0.6.0-py2.7.egg/jwcrypto/jwt.py:`  **in** `def deserialize(self, jwt, key=None):`
```python
        if key is None:
            self.token.deserialize(jwt, None)
        elif isinstance(key, JWK):
            self.token.deserialize(jwt, key)  # <-- Also why it's only deserialized here but not validated? 
        elif isinstance(key, JWKSet):
            self.token.deserialize(jwt, None)
            if 'kid' not in self.token.jose_header:
                raise JWTMissingKeyID('No key ID in JWT header')

            token_key = key.get_key(self.token.jose_header['kid'])
            if not token_key:
                raise JWTMissingKey('Key ID %s not in key set'
                                    % self.token.jose_header['kid'])

            if isinstance(self.token, JWE):
                self.token.decrypt(token_key)
            elif isinstance(self.token, JWS):
                self.token.verify(token_key)
            else:
                raise RuntimeError("Unknown Token Type")
```

How we can deal with it? In case if there is no certain "kid" in a token's header could we perhaps iterate over all available `jwk.JWK` in a `jwk.JWKSet` until for one of them the signature matches. And if neither have worked out then throw an exception. I propose to add this kind of logic in the library. What do you think? 

Currently, a workaround is possible. E.g.
```python
for i, k in enumerate(jwks):
    try:
        jwt = JWT(jwt=token, key=k)
    except Exception as e:
        if (i == len(jwks) - 1):
            raise
        jwt = None

```
But it's not very nice because
1. from the code snippet above-above you can see that if key is just a JWK then no verification happens. I don't quite understand why.
2.  it breaks a nice encapsulation we used to have here. In general our use case looks like the following.
```python
# our use case 
response = requests.get(url)
jwks = JWKSet.from_json(response.text)
jwt = JWT(jwt=token, key=jwks)
claims = json.loads(jwt.claims)
...
```

```cpp
/* It might be that I misunderstood smth, I'm new to the JWT stuff overall */
```

simo5:
Instead of refusing to deserialize a JWT when there is no kid and we are
given a Key Set, try all keys and see if one works.

Fixes #143

### DIFF
--- a/jwcrypto/tests.py
+++ b/jwcrypto/tests.py
@@ -1094,20 +1094,35 @@ class TestJWT(unittest.TestCase):
 
     def test_decrypt_keyset(self):
         key = jwk.JWK(kid='testkey', **E_A2_key)
-        keyset = jwk.JWKSet()
-        # decrypt without keyid
-        t = jwt.JWT(A1_header, A1_claims)
-        t.make_encrypted_token(key)
-        token = t.serialize()
-        self.assertRaises(jwt.JWTMissingKeyID, jwt.JWT, jwt=token,
-                          key=keyset)
-        # encrypt a new JWT
+        keyset = jwk.JWKSet.from_json(json_encode(PrivateKeys))
+
+        # encrypt a new JWT with kid
         header = copy.copy(A1_header)
         header['kid'] = 'testkey'
         t = jwt.JWT(header, A1_claims)
         t.make_encrypted_token(key)
         token = t.serialize()
-        # try to decrypt without key
+        # try to decrypt without a matching key
+        self.assertRaises(jwt.JWTMissingKey, jwt.JWT, jwt=token, key=keyset)
+        # now decrypt with key
+        keyset.add(key)
+        jwt.JWT(jwt=token, key=keyset, check_claims={'exp': 1300819380})
+
+        # encrypt a new JWT with wrong kid
+        header = copy.copy(A1_header)
+        header['kid'] = '1'
+        t = jwt.JWT(header, A1_claims)
+        t.make_encrypted_token(key)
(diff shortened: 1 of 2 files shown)


jwcrypto/tests.py
```diff
@@ -1094,20 +1094,35 @@ def test_A2(self):
...
+
+        keyset = jwk.JWKSet.from_json(json_encode(PrivateKeys))
+        # encrypt a new JWT with no kid
         header = copy.copy(A1_header)
-        header['kid'] = 'testkey'
         t = jwt.JWT(header, A1_claims)
         t.make_encrypted_token(key)
         token = t.serialize()
-        # try to decrypt without key
+        # try to decrypt without key a matching key
```
Esenin: > key a matching key

typo

PR #145, opened 2018-12-24
Add Public and Internal Interfaces

vir-mir:
Dear community `jwcrypto`!

I made changes [PEP8 Public and Internal Interfaces](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces)
 to allow users to extend the library for personal use.

Make a review please.

### DIFF
--- a/docs/source/jwe.rst
+++ b/docs/source/jwe.rst
@@ -17,7 +17,7 @@ Classes
 Variables
 ---------
 
-.. autodata:: jwcrypto.jwe.jwe_default_allowed_algs
+.. autodata:: jwcrypto.jwe.default_allowed_algs
 
 Exceptions
 ----------
--- a/docs/source/jwe.rst
+++ b/docs/source/jwe.rst
@@ -17,7 +17,7 @@ Classes
 Variables
 ---------
 
-.. autodata:: jwcrypto.jwe.default_allowed_algs
+.. autodata:: jwcrypto.jwe.jwe_default_allowed_algs
 
 Exceptions
 ----------
--- a/jwcrypto/__init__.py
+++ b/jwcrypto/__init__.py
@@ -1,0 +1,11 @@
+from jwcrypto.common import *  # noqa: F403 pylint: disable=wildcard-import
+from jwcrypto.jwa import *  # noqa: F403 pylint: disable=wildcard-import
+from jwcrypto.jwk import *  # noqa: F403 pylint: disable=wildcard-import
+from jwcrypto.jws import *  # noqa: F403 pylint: disable=wildcard-import
+from jwcrypto.jwt import *  # noqa: F403 pylint: disable=wildcard-import
+
+__all__ = common.__all__  # noqa: F405 pylint: disable=E0602
+__all__ += jwa.__all__  # noqa: F405 pylint: disable=E0602
+__all__ += jwk.__all__  # noqa: F405 pylint: disable=E0602
+__all__ += jws.__all__  # noqa: F405 pylint: disable=E0602
+__all__ += jwt.__all__  # noqa: F405 pylint: disable=E0602
--- a/jwcrypto/jwa.py
+++ b/jwcrypto/jwa.py
@@ -28,6 +28,12 @@ from jwcrypto.common import base64url_decode, base64url_encode
 from jwcrypto.common import json_decode
 from jwcrypto.jwk import JWK
 
+__all__ = [
+    'JWA',
+    'JWAAlgorithm',
+]
+
+
 # Implements RFC 7518 - JSON Web Algorithms (JWA)
 
 
@@ -37,27 +43,22 @@ class JWAAlgorithm(object):
     @abc.abstractproperty
     def name(self):
         """The algorithm Name"""
-        pass
 
     @abc.abstractproperty
     def description(self):
         """A short description"""
-        pass
 
     @abc.abstractproperty
     def keysize(self):
         """The actual/recommended/minimum key size"""
-        pass
 
     @abc.abstractproperty
     def algorithm_usage_location(self):
         """One of 'alg', 'enc' or 'JWK'"""
-        pass
 
     @abc.abstractproperty
     def algorithm_use(self):
         """One of 'sig', 'kex', 'enc'"""
-        pass
 
 
 def _bitsize(x):
--- a/jwcrypto/jwa.py
+++ b/jwcrypto/jwa.py
@@ -71,7 +71,7 @@ def _inbytes(x):
 
 def _randombits(x):
     if x % 8 != 0:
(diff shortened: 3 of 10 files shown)


docs/source/jwe.rst
```diff
@@ -17,7 +17,7 @@ Classes
 Variables
 ---------
 
-.. autodata:: jwcrypto.jwe.default_allowed_algs
+.. autodata:: jwcrypto.jwe.jwe_default_allowed_algs
```
simo5: why this rename ?

jwcrypto/__init__.py
```diff
@@ -0,0 +1,11 @@
+from jwcrypto.common import *  # noqa: F403 pylint: disable=wildcard-import
+from jwcrypto.jwa import *  # noqa: F403 pylint: disable=wildcard-import
+from jwcrypto.jwk import *  # noqa: F403 pylint: disable=wildcard-import
+from jwcrypto.jws import *  # noqa: F403 pylint: disable=wildcard-import
+from jwcrypto.jwt import *  # noqa: F403 pylint: disable=wildcard-import
+
```
simo5: why are you exporting it all in __init__ ?
this make loading the module costly even when only one function is needed.

jwcrypto/jwa.py
```diff
@@ -28,6 +28,12 @@
 from jwcrypto.common import json_decode
 from jwcrypto.jwk import JWK
 
+__all__ = [
+    'JWA',
+    'JWAAlgorithm',
```
simo5: JWAAlgorithm is not really public ...

jwcrypto/jwe.py
```diff
@@ -8,6 +8,17 @@
 from jwcrypto.common import json_decode, json_encode
 from jwcrypto.jwa import JWA
 
+__all__ = [
+    'JWEHeaderRegistry',
```
simo5: this is not really public

jwcrypto/jwk.py
```diff
@@ -17,6 +17,24 @@
...
+    'JWKTypesRegistry',
+    'ParmType',
+    'JWKParameter',
+    'JWKValuesRegistry',
+    'JWKParamsRegistry',
+    'JWKEllipticCurveRegistry',
+    'JWKUseRegistry',
+    'JWKOperationsRegistry',
+    'JWKpycaCurveMap',
+    'InvalidJWKType',
```
simo5: all of the above are not public intrerfaces

jwcrypto/jws.py
```diff
@@ -8,6 +8,17 @@
 from jwcrypto.jwa import JWA
 from jwcrypto.jwk import JWK
 
+__all__ = [
+    'JWSHeaderParameter',
+    'JWSHeaderRegistry',
+    'jws_default_allowed_algs',
+    'InvalidJWSSignature',
```
simo5: same here also JWSCore is not really public

jwcrypto/jws.py
```diff
@@ -31,7 +42,7 @@
 }
 """Registry of valid header parameters"""
 
-default_allowed_algs = [
+jws_default_allowed_algs = [
```
simo5: why the rename ?
the default_allowed_algs is an accessible property but not a pulic interface

simo5:
Hi @vir-mir are you still interested in pursuing this PR ?

vir-mir:
Hi, @simo5 , yes, but, I need time to finish it. I plan to do this next month.

jwcrypto/jwe.py
vir-mir: Why? if I need to register a new header, for example, `exp` in order to limit the lifetime, and specify it for mandatory processing in the `crit` list.

jwcrypto/jwa.py
vir-mir: What for? `JWAAlgorithm` is an abstract class that allows you to write your own algorithm, defining which methods should be implemented.

jwcrypto/jwk.py
vir-mir: ok, I agree that `JWKUseRegistry`, `JWKOperationsRegistry`, `JWKpycaCurveMap` is not public, but `InvalidJWKType` seems to me to refer to public methods, as it allows to handle exceptions

jwcrypto/jws.py
vir-mir: I agree

jwcrypto/jws.py
vir-mir: I agree, this is a bad decision, but then probably we should pass on yet the here (https://github.com/latchset/jwcrypto/blob/master/jwcrypto/jws.py#L482) parametr `self.allowed_algs` just like here (https://github.com/latchset/jwcrypto/blob/master/jwcrypto/jws.py#L288)

docs/source/jwe.rst
vir-mir: I agree, this is a bad decision

jwcrypto/__init__.py
vir-mir: The import occurs once, usually at the start of the application, and since we have the variable `__all__`, this allows us not to clutter the environment, which is recommended by the python documentation https://docs.python.org/3/tutorial/modules.html#importing-from-a-package

> Remember, there is nothing wrong with using from Package import specific_submodule! In fact, this is the recommended notation unless the importing module needs to use submodules with the same name from different packages.

thus, we hide all the internal logic by implementing the facade design pattern, which makes it easy to find the right class or function for the user and makes it easier to further support and develop library.

jwcrypto/jwa.py
simo5: It is available but should be imported explicitly imo, it is extremely rare (and somewhat discouraged) to implement your own crypto primitieves, and if a new RFC that defines new algorithms come out I rather have the implementation directly in jwcrypto itself.

jwcrypto/jwe.py
simo5: See the PR that landed recently, there is now a mechanism to provide your own custom headers.
In fact you need to rebase the PR as that one breaks yours.

jwcrypto/jwk.py
simo5: ok for InvalidJWKType

jwcrypto/jwa.py
vir-mir: I agree. I will remove `JWAAlgorithm`

jwcrypto/jwe.py
vir-mir: You are talking about https://github.com/latchset/jwcrypto/pull/148
Yes, now it is enough `JWSEHeaderParameter` from the module `common`

simo5:
Closing as stale, please feel free to reopen or create a new PR in the future if you sill want to contribute it.

Issue #146, opened 2019-03-07
Populating x5u/x5c fields


lumjjb:
Is there currently a way to populate `x5u/x5c` fields? 

Would there be interest if we could provide a PR for the feature if not currently possible?

simo5:
JWK, JWS and JWE classes do know about the existence of these parameters, but they do not know how to handle them, so if they are marked critical the implementation will raise an error.
However if they are not critical they will be retained as is.

Implementing support for directly downloading certificates (x5u) or validating the certificate chain (x5c) would not be simple as I do not want to force dependencies on additional python modules and also I do not want to perform network operations behind the back of the main application.

However there is a todo in the code about support custom headers so that an application can make use of them. What we'd need is probably a way to register new entries in JWSHeaderRegistry via some setter/class function and also have a way to provide a list of validation functions/objects for custom handling.

Once we have that than we can also have a module that implements basic URL and Chain validation handling that applications can use or override as they need.

I can help steering the design if you want to try to come up with a proposal.

lumjjb:
>However there is a todo in the code about support custom headers so that an application can make use of them. What we'd need is probably a way to register new entries in JWSHeaderRegistry via some setter/class function and also have a way to provide a list of validation functions/objects for custom handling.

Thanks @simo5, that sounds perfect! Understand the concern of dependencies in this respect. This in itself would be suitable for our purposes, we have an external component that does the validation, so all we need is a way to add the entry to `JWSHeaderRegistry`.

>I can help steering the design if you want to try to come up with a proposal.

We are finishing up some other existing work, but this would be the next thing we need to look at, so we will share with you what we have in mind for that. :)

simo5:
ack

lumjjb:
I was looking through the library, and it seems like the safest way to add headers would be to use the existing `_merge_check_headers` function. I noticed that registry check was not in the function itself, so added it as a wrapper:

For `jwcrypto/jws.py`:
```
    def add_headers (self, *headers):
        for hn in headers:
            if hn is None:
                continue
            for h in list(hn.keys()):
                if h not in JWSHeaderRegistry:
                    raise InvalidJWSObject('Header not valid: "%s"' % h)

        self.objects['protected'] = json_encode(self._merge_check_headers (json_decode(self.objects['protected']), *headers))
```

I am thinking to instead move the check for `if h not in JWSHeaderRegistry:` into `_merge_check_headers`. 

[...]

PR #148, opened 2019-03-15
Create extendable HeaderRegistry 

lumjjb:
Create a way to extend the functionality with custom header checking without adding dependencies into the main library, as discussed in #146 

Current implementation only takes in `JWSEHeaderParameter`, but it can be extendable to any kind of header parameter (like JWK) if we take in a set of validation functions that tells us if a header can be ovewritten, and how to reference the `check_fn`. Would like to get some feedback on this.

Side note: Took the opportunity to merge the HeaderParameter types of JWE and JWS (according to RFC 7515 - 9.1).

### DIFF
--- a/jwcrypto/common.py
+++ b/jwcrypto/common.py
@@ -1,8 +1,10 @@
 # Copyright (C) 2015 JWCrypto Project Contributors - see LICENSE file
 
+import copy
 import json
 from base64 import urlsafe_b64decode, urlsafe_b64encode
-
+from collections import MutableMapping
+from collections import namedtuple
 
 # Padding stripping versions as described in
 # RFC 7515 Appendix C
@@ -104,3 +106,85 @@ class InvalidJWEKeyLength(JWException):
     def __init__(self, expected, obtained):
         msg = 'Expected key of length %d, got %d' % (expected, obtained)
         super(InvalidJWEKeyLength, self).__init__(msg)
+
+
+class InvalidJWSERegOperation(JWException):
+    """Invalid JWSE Header Registry Operation.
+
+    This exception is raised when there is an error in trying ot add a JW
+    Signature or Encryption header to the Registry.
+    """
+
+    def __init__(self, message=None, exception=None):
+        msg = None
+        if message:
+            msg = message
+        else:
+            msg = 'Unknown Operation Failure'
+        if exception:
+            msg += ' {%s}' % repr(exception)
+        super(InvalidJWSERegOperation, self).__init__(msg)
+
+
+# JWSE Header Registry definitions
+
+# RFC 7515 - 9.1: JSON Web Signature and Encryption Header Parameters Registry
+# HeaderParameters are for both JWS and JWE
+JWSEHeaderParameter = namedtuple('Parameter',
+                                 'description mustprotect supported check_fn')
+
+
+class JWSEHeaderRegistry(MutableMapping):
+    def __init__(self, init_registry=None):
+        if init_registry:
+            if isinstance(init_registry, dict):
+                self._registry = copy.deepcopy(init_registry)
+            else:
+                raise InvalidJWSERegOperation('Unknown input type')
+        else:
+            self._registry = {}
+
+        MutableMapping.__init__(self)
+
+    def check_header(self, h, value):
+        if h not in self._registry:
+            raise InvalidJWSERegOperation('No header "%s" found in registry'
+                                          % h)
+
+        param = self._registry[h]
+        if param.check_fn is None:
+            return True
+        else:
+            return param.check_fn(value)
+
+    def __getitem__(self, key):
+        return self._registry.__getitem__(key)
+
+    def __iter__(self):
+        return self._registry.__iter__()
+
+    def __delitem__(self, key):
+        if self._registry[key].mustprotect or \
+           self._registry[key].supported:
+            raise InvalidJWSERegOperation('Unable to delete protected or '
+                                          'supported field')
+        else:
+            self._registry.__delitem__(key)
+
+    def __setitem__(self, h, jwse_header_param):
+        # Check if a header is not supported
+        if h in self._registry:
+            p = self._registry[h]
+            if p.supported:
+                raise InvalidJWSERegOperation('Supported header already exists'
+                                              ' in registry')
+            elif p.mustprotect and not jwse_header_param.mustprotect:
+                raise InvalidJWSERegOperation('Header specified should be'
+                                              'a protected header')
+            else:
+                del self._registry[h]
+
+        self._registry[h] = jwse_header_param
+
+    def __len__(self):
+        return self._registry.__len__()
--- a/jwcrypto/jwe.py
+++ b/jwcrypto/jwe.py
@@ -4,6 +4,7 @@ import zlib
 
 from jwcrypto import common
 from jwcrypto.common import JWException
+from jwcrypto.common import JWSEHeaderParameter, JWSEHeaderRegistry
 from jwcrypto.common import base64url_decode, base64url_encode
 from jwcrypto.common import json_decode, json_encode
 from jwcrypto.jwa import JWA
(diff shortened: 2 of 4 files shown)


lumjjb:
For local try-out, i used:

```
# Test check 1

# Check function takes in a JWS and checks if it's custom1 header == 'custom-text'
def check_custom1(jwobj):
    return jwobj.jose_header['custom1'] == 'custom-text'

Token = jwt.JWT(header={"alg": "RS256", 'custom1':'custom-text'}, claims={"claim1": "claim1value"})
Token.make_signed_token(key)
Token.token.header_registry['custom1'] = JWSEHeaderParameter('Custom header 1', False, True, check_custom1)

assert (Token.token.header_registry.check_header('custom1', Token.token))

Token = jwt.JWT(header={"alg": "RS256", 'custom1':'not-custom-text'}, claims={"claim1": "claim1value"})
Token.make_signed_token(key)
Token.token.header_registry['custom1'] = JWSEHeaderParameter('Custom header 1', False, True, check_custom1)

assert (not Token.token.header_registry.check_header('custom1', Token.token))
```

jwcrypto/jwe.py
```diff
@@ -89,6 +90,7 @@ def __init__(self, plaintext=None, protected=None, unprotected=None,
         self._allowed_algs = None
         self.objects = dict()
         self.plaintext = None
+        self.header_registry = JWSEHeaderRegistry(DefaultJWEHeaderRegistry)
```
simo5: Why the rewrap here ?

simo5: Or maybe why DefaultJWEHeaderRegistry is a JWSEHeaderRegistry() itself instead of being just a dict ?

simo5:
It would be nice to have a function or init() argument to allow to create a JWS/JWE and pass in custom headers to be accepted.
Also at least one unit test per JWS and JWE with a custom header.

jwcrypto/jwe.py
lumjjb: One reason I thought about this was in the current use, the header functions would have to be registered separately on each instantiation. I was thinking in the case where we had in mind to add functionality for a particular header throughout the program, in that case, we would add the header to the `DefaultRegistry`, and every object `JWS/JWE` created from then on would have that functionality built it.

lumjjb:
> It would be nice to have a function or init() argument to allow to create a JWS/JWE and pass in custom headers to be accepted.

Got it, will add an init argument here.

> Also at least one unit test per JWS and JWE with a custom header.

Gotcha! Btw - i had a quick question about the naming of the test variables, are these particularly named to follow any examples in the spec? Or are they just arbitrary and I can create a new `A8_example`?

jwcrypto/jwe.py
simo5: uhmm, changing the DefaultJWEHeaderRegistry would make everything brittle, because i would change how every caller works, including callers in other libraries.
Please do not do that.

By allowing to pass a new registry via an init parameter in the JWE you can create your own registry once, and then always pass that one into your own controlled JWEs.

This allow the "cost savings" you want without messing up static global objects.

so I think let's get rid of DefaultJWEHeaderRegistry, and maintain only the dictionary, and move initialization in JWE/JWS constructors, including the ability to use a passed in Registry.

simo5:
They are names after the RFC Appendix, if you need to add new tests not covered in the RFC Appendix please chose a different prefix.

jwcrypto/jwe.py
lumjjb: Yea - that makes sense. I will rename it back to `JWEHeaderRegistry` as well then so it doesn't break any libraries referencing that particular variable.

jwcrypto/common.py
```diff
@@ -104,3 +106,83 @@ class InvalidJWEKeyLength(JWException):
...
+# HeaderParameters are for both JWS and JWE
+JWSEHeaderParameter = namedtuple('Parameter',
+                                'description mustprotect supported check_fn')
+
+class JWSEHeaderRegistry:
+    def __init__(self, init_registry=None):
+        if init_registry:
+            if isinstance(init_registry, dict):
+                self._registry = copy.deepcopy(init_registry)
+            elif isinstance(init_registry, JWSEHeaderRegistry):
```
simo5: I think we can drop this now ?

jwcrypto/common.py
```diff
@@ -104,3 +106,83 @@ class InvalidJWEKeyLength(JWException):
...
+
+
+# JWSE Header Registry definitions
+
+# RFC 7515 - 9.1: JSON Web Signature and Encryption Header Parameters Registry
+# HeaderParameters are for both JWS and JWE
+JWSEHeaderParameter = namedtuple('Parameter',
+                                'description mustprotect supported check_fn')
+
+class JWSEHeaderRegistry:
```
simo5: shouldn't this simply extend 'dict' so that you do not ahve to redefine a dict-like interface, but only re-implement the __setitem__/__delitem__ calls ?

jwcrypto/tests.py
```diff
@@ -679,6 +694,27 @@ def test_E(self):
...
 
+    def test_customhdr_jwe(self):
+        s = jws.JWS(A6_example['payload'])
+        s.deserialize(customhdr_jws_example, A6_example['key2'])
+
+        def check_custom1(jwobj):
+            return jwobj.jose_header['custom1'] == 'custom_val'
+
+        s.header_registry['custom1'] = JWSEHeaderParameter('Custom header 1', False, True, check_custom1)
+        assert (s.header_registry.check_header('custom1', s))
```
simo5: For this and following tests ... they kinda work as a sort of unit test but are not testing the integration into the JWS and JWE machinery.
I would expect to see tests that try to create/verify a JWS/JWE and see that with the incorrect/missing custom header we get an exception.

jwcrypto/jwe.py
```diff
@@ -91,7 +92,10 @@ def __init__(self, plaintext=None, protected=None, unprotected=None,
         self._allowed_algs = None
         self.objects = dict()
         self.plaintext = None
-        self.header_registry = JWSEHeaderRegistry(header_registry)
+        if header_registry:
+            self.header_registry = JWSEHeaderRegistry(header_registry)
+        else:
+            self.header_registry = JWSEHeaderRegistry(JWEHeaderRegistry)
```
simo5: Why this change ?

jwcrypto/common.py
lumjjb: Yea we can drop this now.

jwcrypto/common.py
lumjjb: I was hesitant about extending the `dict` class directly due to some dangers of weird built-in implementation specific issues. One of the few were:

http://www.kr41.net/2016/03-23-dont_inherit_python_builtin_dict_type.html

What do you think about this? I will defer that decision to you, I am okay with either but personal preference of not extending the `dict` class directly.

jwcrypto/tests.py
lumjjb: I didn't integrate the `check_header` into the actual signature verification and decryption process. But based on this comment, I think that's something that you'd want yea?

jwcrypto/common.py
simo5: Uhmm  ... interesting read.
Maybe use the Mutable thing instead ?
We already import namedtuple from collections, so that is ok to have as a dependency I guess.

jwcrypto/tests.py
simo5: Yea, that's the way people would expect tohave it working I think.

jwcrypto/common.py
lumjjb: Ok sg, i'll use that then.

jwcrypto/tests.py
lumjjb: Alright, let me add that.

jwcrypto/jwe.py
lumjjb: `pep8` complained about a dictionary as a default value... Since it's mutable, and unfortunately, there isn't a frozen version of dictionary that it recognizes afaik.

lumjjb:
Made the changes. Let me know if all is good and i'll squash.

jwcrypto/jws.py
```diff
@@ -264,7 +264,12 @@ def _verify(self, alg, key, payload, signature, protected, header=None):
                 raise InvalidJWSSignature('Invalid Unprotected header')
 
         # Merge and check (critical) headers
-        self._merge_check_headers(p, header)
+        chk_hdrs = self._merge_check_headers(p, header)
+        for hdr in chk_hdrs:
```
simo5: I would expect headers checks to happen inside the function named _merge_check_headers() ?

jwcrypto/tests.py
```diff
@@ -695,27 +695,26 @@ def test_E(self):
             s.verify(None)
 
     def test_customhdr_jws(self):
+        # Test pass header check
         s = jws.JWS(A6_example['payload'])
```
simo5: I would expect JWS to be initialized with the dictionary that defines a custom header check ?

jwcrypto/tests.py
```diff
@@ -988,25 +987,23 @@ def test_JWE_Issue_136(self):
 
     def test_customhdr_jwe(self):
         e = jwe.JWE()
```
simo5: as well here JWE init should get the custom header registry ?

jwcrypto/jwe.py
```diff
@@ -492,7 +506,7 @@ def payload(self):
 
     @property
     def jose_header(self):
-        jh = self._get_jose_header()
+        jh = self._get_jose_header(self.objects.get('header'))
```
simo5: This is a good catch, but would be nice to have it in a separate PR I can push right away,

lumjjb: Ok sounds good, I will create a separate PR for this!

jwcrypto/tests.py
lumjjb: Got it, i added the header later on, but agree would be fitting to have it as a custom registry.

jwcrypto/jws.py
lumjjb: yea sounds good, will try that.

lumjjb: This doesn't work well when I tried it. What ends up happening is that like in the header extension check function that i created in the test, which uses `jose_header`, it calls `self._merge_check_headers`, which ends up in a recursive loops.

simo5: merge and check used to be two separate functions,
let's leave the code as you have it now with a FIXME comment, and I'll take care of restructuring at a later time.

lumjjb:
- Made the req. changes 
- made it so that only allow adding to reg
- added test for dup headers

jwcrypto/jwe.py
```diff
@@ -354,6 +363,11 @@ def _decrypt(self, key, ppe):
         # TODO: allow caller to specify list of headers it understands
```
simo5: Time to drop this TODO :-D

jwcrypto/jws.py
```diff
@@ -213,6 +211,7 @@ def is_valid(self):
         return self.objects.get('valid', False)
 
     # TODO: allow caller to specify list of headers it understands
```
simo5: and this one too

simo5:
Thanks a lot for this contribution!

PR #155, opened 2019-04-03
RFC 8037 - Support for Ed25519, Ed448

andrewmackie:
Support for Ed25519 and Ed448, allowing signing and verification.

Some code cherry-picked from [Simo5's rfc8037 branch](https://github.com/simo5/jwcrypto/tree/rfc8037).

### DIFF
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -9,6 +9,8 @@ from cryptography import x509
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.hazmat.primitives.asymmetric import ed25519
+from cryptography.hazmat.primitives.asymmetric import ed448
 from cryptography.hazmat.primitives.asymmetric import rsa
 
 from six import iteritems
@@ -18,10 +20,11 @@ from jwcrypto.common import base64url_decode, base64url_encode
 from jwcrypto.common import json_decode, json_encode
 
 
-# RFC 7518 - 7.4
+# RFC 7518 - 7.4 , RFC 8037 - 5
 JWKTypesRegistry = {'EC': 'Elliptic Curve',
                     'RSA': 'RSA',
-                    'oct': 'Octet sequence'}
+                    'oct': 'Octet sequence',
+                    'OKP': 'Octet Key Pair'}
 """Registry of valid Key Types"""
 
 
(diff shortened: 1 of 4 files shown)


jwcrypto/jwk.py
```diff
@@ -79,10 +87,12 @@ class ParmType(Enum):
 }
 """Regstry of valid key parameters"""
 
-# RFC 7518 - 7.6
+# RFC 7518 - 7.6 , RFC 8037 - 5
 JWKEllipticCurveRegistry = {'P-256': 'P-256 curve',
                             'P-384': 'P-384 curve',
-                            'P-521': 'P-521 curve'}
+                            'P-521': 'P-521 curve',
+                            'Ed25519': 'Ed25519 signature algorithm key pairs',
```
simo5: This should probably just be: "Ed25519 curve"

simo5: Or maybe better: "Edwards-curve 25519 for Digital Signatures"

jwcrypto/jwk.py
```diff
@@ -353,6 +366,49 @@ def _import_pyca_pub_ec(self, key, **params):
         )
         self.import_key(**params)
 
+    def _generate_OKP(self, params):
+        curve = 'Ed25519'  # Ed25519 is the default Edwards curve
+        if 'curve' in params:
+            curve = params.pop('curve')
+        # 'curve' is for backwards compat, if 'crv' is defined it takes
```
simo5: given this is a new type please let's drop the old "curve" parameter support and use exclusively "crv"

jwcrypto/tests.py
```diff
@@ -19,7 +19,7 @@
 from jwcrypto.common import base64url_decode, base64url_encode
 from jwcrypto.common import json_decode, json_encode
 
-# RFC 7517 - A.1
+# RFC 7517 - A1
```
simo5: leave A.1 please

jwcrypto/jwk.py
andrewmackie: Thanks - I took that directly from section 5 of RFC 8037:

```
   The following has been added to the "JSON Web Key Elliptic Curve"
   registry:

   o  Curve Name: "Ed25519"
   o  Curve Description: Ed25519 signature algorithm key pairs
```
Am happy either way.

jwcrypto/jwk.py
andrewmackie: Agreed, will do.

jwcrypto/tests.py
andrewmackie: sorry, no problem.

andrewmackie:
I've pushed two of the changes you requested - let me know if you still want the curve description changed (see my comment above). Then I'll squash the last two commits into one for you.

jwcrypto/jwk.py
simo5: Ah, I think the RFC wording is a bit unfortunate, but ok

andrewmackie:
There it is. Thanks for your patience with me, I really appreciate it (and thanks for jwcrypto too - it's great).

simo5:
Thank you for the contirbution, I really appreciated it!

PR #160, opened 2019-08-07
typo correction

pmarillo:


simo5:
Uhm the failure is unrelated but concerning.
@tiran could you take a look ?
Meanwhile I'll push this PR

PR #164, opened 2019-09-27
Add the ability to verify 'none' signatures.

### LINKED ISSUE #163: Support for plaintext JWT
Is there a plan to support unpacking of [plaintext JWTs](https://openid.net/specs/draft-jones-json-web-token-07.html#Plaintext)?

Currently, I have to handle them manually, since it is not possible to access token data before calling verify, and `verify` fails when no signature is present, although `alg` is specified to be `none`.

simo5:
Apparently this is required by OpenID Connect as an option:
https://openid.net/specs/openid-connect-core-1_0.html#RequestObject

In order to avoid accidental use of the None algoritm, which is disabled
by default but easy to casually enable, add the requirement to pass a
"None Key".
The none key clearly indicates that the user really wants to allow the
'none' algorithm to be used.

The "None Key" is an 'oct' key with key size of 0.

Example:
```
>>> from jwcrypto import jwk,jwt
>>> e="eyJhbGciOiJub25lIn0.eyJpc3MiOiJqb2UiLCJodHRwOi8vZXhhbXBsZS5jb20vaXNfcm9vdCI6dHJ1ZX0."
>>> k=jwk.JWK(generate='oct', size=0)
>>> E=jwt.JWT(algs=['none'])
>>> E.deserialize(jwt=e, key=k)
>>> E.claims
u'{"iss":"joe","http://example.com/is_root":true}'
```

Fixes #163

### DIFF
--- a/jwcrypto/jwa.py
+++ b/jwcrypto/jwa.py
@@ -161,7 +161,8 @@ class _RawNone(_RawJWS):
         return ''
 
     def verify(self, key, payload, signature):
-        raise InvalidSignature('The "none" signature cannot be verified')
+        if key.key_type != 'oct' or key.get_op_key() != '':
+            raise InvalidSignature('The "none" signature cannot be verified')
 
 
 class _HS256(_RawHMAC, JWAAlgorithm):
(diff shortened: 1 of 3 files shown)


simo5:
Ok, fixed current tests and adding a new one.

PR #168, opened 2020-02-01
Import ABC from collections.abc instead of collections for Python 3.9 compatibility

### LINKED ISSUE #167: Importing ABC directly from collections was removed in Python 3.9
https://github.com/latchset/jwcrypto/blob/58c1099e531546feb473ad6bd34f9cbba7040d0d/jwcrypto/common.py#L6

tirkarthi:
Fixes #167


simo5:
Unclear to me why sphinx is failing but I do not see how it could be related

PR #173, opened 2020-03-19
Add secp256k1 EC Key curve support

synkc:
This is as per
- https://tools.ietf.org/html/draft-ietf-cose-webauthn-algorithms
- https://mailarchive.ietf.org/arch/msg/cose/91MouVA43DefbpQOB7l5daCEeIc

### DIFF
--- a/jwcrypto/jws.py
+++ b/jwcrypto/jws.py
@@ -480,7 +480,9 @@ class JWS(object):
         if alg is None:
             raise ValueError('"alg" not specified')
 
-        c = JWSCore(alg, key, protected, self.objects['payload'])
+        c = JWSCore(
+            alg, key, protected, self.objects['payload'], self.allowed_algs
+        )
         sig = c.sign()
 
         o = dict()
--- a/jwcrypto/tests.py
+++ b/jwcrypto/tests.py
@@ -259,6 +259,29 @@ PrivateKeys_EdDsa = {
     ]
 }
 
+PublicKeys_secp256k1 = {
+    "keys": [
+        {
+            "kty": "EC",
+            "crv": "secp256k1",
+            "x": "Ss6na3mcci8Ud4lQrjaB_T40sfKApEcl2RLIWOJdjow",
+            "y": "7l9qIKtKPW6oEiOYBt7r22Sm0mtFJU-yBkkvMvpscd8"
+        }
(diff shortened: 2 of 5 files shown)


jwcrypto/jws.py
```diff
@@ -30,7 +30,7 @@
     'RS256', 'RS384', 'RS512',
     'ES256', 'ES384', 'ES512',
     'PS256', 'PS384', 'PS512',
-    'EdDSA']
+    'EdDSA', 'ES256K']
```
simo5: I think we shouldn't allow this by default until the draft is converted to an RFC.

jwcrypto/tests.py
```diff
@@ -786,6 +822,17 @@ def test_EdDsa_signing_and_verification(self):
...
+        key = jwk.JWK(**PrivateKeys_secp256k1['keys'][0])
+        payload = bytes(bytearray(A1_payload))
+        jws_test = jws.JWS(payload)
+        jws_test.add_signature(key, None, json_encode({"alg": "ES256K"}), None)
+        jws_test_serialization_compact = jws_test.serialize(compact=True)
+        jws_verify = jws.JWS()
+        jws_verify.deserialize(jws_test_serialization_compact)
+        jws_verify.verify(key.public())
+        self.assertEqual(jws_verify.payload, payload)
+
```
simo5: This is good, but only guarantee we interoperate with ourselves.
Unfortunately I see no test vectors in the proposed draft, that should be addressed at IETF sending a comment they should provide official test vectors for both the secp256k1 curve as well as the RSASSA-PKCS1-5 algorithms.

jwcrypto/jws.py
synkc: Fixed to fba5603.

jwcrypto/tests.py
synkc: What do you think about merging this PR until RFC comes out? If the test vector comes out later, it would be better to modify it.

simo5: Yes I am ok merging, this is why I asked to remove from default list.
Once it is final we can make any modification that is needed and add it to the default list.

Could you please squash the commits?
I will then merge.

synkc:
Squashed. thank you!

PR #177, opened 2020-04-19
Error importing OKP private key

### LINKED ISSUE #176: Error importing OKP private key
You can see the problem in this stack trace:

```
  File "/home/david/python-jwt/test/fixtures.py", line 136, in <module>
    okp_ed25519_priv_key = JWK.from_pem(to_bytes_2and3(okp_ed25519_priv_pem))
  File "/home/david/.local/lib/python2.7/site-packages/jwcrypto/jwk.py", line 904, in from_pem
    obj.import_from_pem(data, password)
  File "/home/david/.local/lib/python2.7/site-packages/jwcrypto/jwk.py", line 851, in import_from_pem
    self.import_from_pyca(key)
  File "/home/david/.local/lib/python2.7/site-packages/jwcrypto/jwk.py", line 818, in import_from_pyca
    self._import_pyca_pri_okp(key)
  File "/home/david/.local/lib/python2.7/site-packages/jwcrypto/jwk.py", line 456, in _import_pyca_pri_okp
    crv=params['crv'],
KeyError: 'crv'
```
`_import_pyca_pri_okp` expects `params` to contain `crv`: https://github.com/latchset/jwcrypto/blob/master/jwcrypto/jwk.py#L461

However, when importing, no `params` are passed down: https://github.com/latchset/jwcrypto/blob/master/jwcrypto/jwk.py#L823

eirnym:
Fixes #176

### DIFF
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -810,19 +810,19 @@ class JWK(object):
         else:
             raise NotImplementedError
 
-    def import_from_pyca(self, key):
+    def import_from_pyca(self, key, **params):
         if isinstance(key, rsa.RSAPrivateKey):
-            self._import_pyca_pri_rsa(key)
+            self._import_pyca_pri_rsa(key, **params)
         elif isinstance(key, rsa.RSAPublicKey):
-            self._import_pyca_pub_rsa(key)
+            self._import_pyca_pub_rsa(key, **params)
         elif isinstance(key, ec.EllipticCurvePrivateKey):
-            self._import_pyca_pri_ec(key)
+            self._import_pyca_pri_ec(key, **params)
         elif isinstance(key, ec.EllipticCurvePublicKey):
-            self._import_pyca_pub_ec(key)
+            self._import_pyca_pub_ec(key, **params)
         elif isinstance(key, (Ed25519PrivateKey, Ed448PrivateKey)):
-            self._import_pyca_pri_okp(key)
+            self._import_pyca_pri_okp(key, **params)
         elif isinstance(key, (Ed25519PublicKey, Ed448PublicKey)):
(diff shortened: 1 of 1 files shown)


simo5:
I am not gainst this patch in principle, but it would be an enhancement and not really a fix as the interface is supposed to work w/o any additional param.

That said, I also wonder what kind of params do you see being passed in here ?
If there is a legitimate use I will be happy to merge this PR (with a test though)

eirnym:
In this case I wonder why these functions all accept params in the first place.

Functions such as `_generate_RSA` push their params, while most params are taken from the key itself. 

This'd make code more consistent to either add `**params` or remove them everywhere.

simo5:
With a "generate" function you (as the user of the API) may want to add for example a 'kid' parameter to give the key a specific name.
I do not know if we want to do that with PEM imports, but if we do then yes we should pass parameters there too.
But we need to also define what happens if there is a conflict between the parameters source from the PEM file and those passed in, and that's where it becomes complicated :-)

eirnym:
Some parameters are overwritten by corresponding function in any case. 

In my opinion, if we generate and import we should pass or not to pass `**params` everywhere.

E.g. on PEM example, `_generate_OKP` generates a key and importing it immediately after, passing `**params`. (The same is true for other `_generate_*` functions). 

The API consistency is very important

simo5:
I am not too worried by that level of consistency, as consistency is good only if there is the same intent behind an API.
We also cannot remove exiting functionality now so we have only 2 options:
1. add **params to pem import functions, but then we need to also define what teh semantics are, which cannot be the same as generate functions
2. Do not add **params to pem import functions, and then things stay as they are.

eirnym:
Now I see your point, thank you for a detailed explanation

PR #180, opened 2020-04-22
Add method to export Keys ans Sets as dictionaries

### LINKED ISSUE #28: Import/Export consistency
The export() methods should return a python dictionary that can be passed to json.dumpf(). Likewise, the import() methods should take a python dictionary as returned from json.loadf(). Please don't do the encoding internal to the API.

simo5:
This allows to avoid json_decoding again to manipulate the exported
data.

Fixes #28

### DIFF
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -576,7 +576,7 @@ class JWK(object):
         obj.import_key(**jkey)
         return obj
 
-    def export(self, private_key=True):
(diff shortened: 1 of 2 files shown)


npmccallum:
Makes sense to me. Although, one might also wonder why not offload the json serialization/deserialization entirely and always return a dict. I suspect the answer is backwards compatibility.

simo5:
Yeah backwars compat is the answer.

PR #188, opened 2020-08-24
Make sure an empty dict is a valid JWT payload

### LINKED ISSUE #187: Empty claims raises error on make_signed_token
If one creates a JWT with an empty claims (`{}`), `make_signed_token` errors out as it gets a claims of `None`.

This would appear to be due to `JWT.__init__` setting `self.claims` on if `claims` evaluates to false, as opposed to on `if claims is not None`.

Should that be changed, or was this on purpose?  (I don't notice in the spec that claims *must* have any key/values specified; I'd expect empty claims to be valid.)

simo5:
Nowhere in the spec it says the payload must not be an empty dictionary.
It makes little sense to not have any claims, because then, what do you
claim? But that's a problem for the user, especially given the fact you
could already pas in the literal string '{}' and it would get signed.

However a completely empty string is invalid, at least one character,
even just a space is needed.

Fixes #187

### DIFF
--- a/jwcrypto/jwt.py
+++ b/jwcrypto/jwt.py
@@ -199,7 +199,7 @@ class JWT(object):
         if check_claims is not None:
             self._check_claims = check_claims
 
-        if claims:
+        if claims is not None:
             self.claims = claims
 
         if jwt is not None:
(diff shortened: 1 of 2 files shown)


simo5:
Reporter says this work for them in #187 and the change has tests, so I am going to forgo further review and push.

PR #189, opened 2020-08-31
RFC 8812 - Add Default allowed algorithms

synkc:
Add ES256K to the Default allowed algorithm
- https://www.rfc-editor.org/rfc/rfc8812.html


simo5:
LGTM

PR #181, opened 2020-04-28
Turn JWK into a dict-like object

simo5:
This PR changes JWK to inherit from dict, and behave like a dictionary

Fixes #27

PR #192, opened 2020-12-03
Adding Power support(ppc64le) with ci and testing to the project for architecture independent

asellappen:
I am working for IBM to port cpu arch ppc64le for open sources.

This PR adds CI support for the IBM Power Little Endian (ppc64le) architecture. The idea is to ensure that the builds on this architecture are continuously tested along with the Intel builds (amd64) as this is part of the ubuntu distro 
on that architecture as well and detecting (and fixing) any issues or failures early would help to ensure that we are always up to date.

This helps us simplify testing later when distributions are re-building and re-releasing,We typically build applications for customers and 
ISVs, and while we don't use this package directly,we do count on all of the packages in debian/ubuntu to build other packages. So we more likely have this as a second or 
third level dependency and couldn't tell you explicitly which features we use or our usage model,For more info tag @gerrith3.

Please help to verify and merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 sudo: false
-
 language: python
 
 cache: pip
@@ -24,7 +23,16 @@ matrix:
       env: TOXENV=pep8py2
     - python: 3.6
       env: TOXENV=pep8py3
+    - python: 3.6
+      arch: ppc64le
+      env: TOXENV=py36
+      dist: Focal
+    - python: 3.8
+      env: TOXENV=py38
+      arch: ppc64le
+      dist: Focal
 
+      
 install:
   - pip install --upgrade pip setuptools
   - pip --version


simo5:
Sounds good to me,
thanks!

simo5:
Can you please squash the commits?

asellappen:
> Can you please squash the commits?

Sorry missed that ,now it is perfect ,Please have a look and merge.

asellappen:
@simo5 ,done thanks .

simo5:
Actually lloking at it we really do not need the shpinx, lint and pep8  taks repeated on a different architecture, and they unnecessarily raise the resources and time needed to run the CI jobs, can you drop them and leave only standard jobs?

Also why do you use py34 TOXENV with non python3.4 jobs?

simo5:
Also please note that I am dropping py34 completely as it doesn't work on Fedora33
PR #181 will do that

asellappen:
@simo5 ,thanks for feed back.
I have removed  shpinx, lint and pep8 tasks and included two standard jobs for ppc64le  , py36 and py38 .
Please verify and approve the changes.

simo5:
The rest of the configurations for travis do not require a specific distro, is there a reason to require Focal specifically ?
I try to avoid it so that as travis updates their defaults the files keep working ...

asellappen:
> The rest of the configurations for travis do not require a specific distro, is there a reason to require Focal specifically ?
> I try to avoid it so that as travis updates their defaults the files keep working ...

during the run of tox on ppc64le,We have faced problem to get support for OpenSSL, It is supported from focal dist onwards, hence i replaced dist for ppc64le from xenial to focal.

simo5:
Alright, makes sense!

asellappen:
thank you .

PR #195, opened 2020-12-11
Deprecate RSA1_5 and remove from defaults

### LINKED ISSUE #193: The pyca/cryptography RSA PKCS#1 v1.5 is unsafe, making the users of it vulnerable to Bleichenbacher attacks
The API provided by pyca/cryptography is not secure, as documented in their docs:
https://github.com/pyca/cryptography/commit/8686d524b7b890bcbe6132b774bd72a3ae37cf0d

As far as I can tell, it's one of the APIs that jwcrypto uses to implement the key management: https://github.com/latchset/jwcrypto/blob/3b7e8322d37a6b07f862d46e6a43b139cbcc89c5/jwcrypto/jwe.py#L36

That would mean that jwcrypto is vulnerable to Bleichenbacher attacks.
To fix it, implementing an API with built-in workarounding of padding oracle attacks is necessary in pyca/cryptography, as briefly described in https://github.com/pyca/cryptography/issues/5510.

simo5:
RSA1_5 relies on the PKCS1 v 1.5 padding algorithm.

This algorithm has been shown time and again to be very difficult to
implement correctly because it requires a fully constant time
implementation including in application level error handling.

Python Cryptography which is our crypto engine documents that the
current API is not contacnt time.

This patch deprecates the algorithm and removes it from the default
allowed algorithms list. This means that code need to be changed in
order to explicitly allow RSA1_5 going forward.

Resolves #193

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,3 +16,23 @@ Documentation
 =============
 
 http://jwcrypto.readthedocs.org
+
+Deprecation Notices
+===================
+
+2020.12.11: The RSA1_5 algorithm is now considered deprecated due to numerous
+implementation issues that make it a very problematic tool to use safely.
+The algorithm can still be used but requires explicitly allowing it on object
+instantiation. If your application depends on it there are examples of how to
+re-enable RSA1_5 usage in the tests files.
+
+Note: if you enable support for `RSA1_5` and the attacker can send you chosen
+ciphertext and is able to measure the processing times of your application,
+then your application will be vulnerable to a Bleichenbacher RSA padding
+oracle, allowing the so-called "Million messages attack". That attack allows
+to decrypt intercepted messages (even if they were encrypted with RSA-OAEP) or
+forge signatures (both RSA-PKCS#1 v1.5 and RSASSA-PSS).
+
+Given JWT is generally used in tokens to sign authorization assertions or to
+encrypt private key material, this is a particularly severe issue, and must
+not be underestimated.
--- a/jwcrypto/jwe.py
+++ b/jwcrypto/jwe.py
@@ -33,7 +33,7 @@ JWEHeaderRegistry = {
 
 default_allowed_algs = [
     # Key Management Algorithms
-    'RSA1_5', 'RSA-OAEP', 'RSA-OAEP-256',
+    'RSA-OAEP', 'RSA-OAEP-256',
     'A128KW', 'A192KW', 'A256KW',
     'dir',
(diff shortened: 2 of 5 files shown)


simo5:
@tomato42 can you take a look at this one ?

README.md
```diff
@@ -16,3 +16,13 @@ Documentation
...
+
+Deprecation Notices
+===================
+
+2020.12.11: The RSA1_5 algorithm is now considered deprecated due to
+numerous implementation issues that make it a very problematic tool to
+use safely.
+The algorithm can still be used but requires explicitly allowing it on
+object instantiation. If your application depends on it there are examples
+of how to re-enable RSA1_5 usage in the tests files.
```
tomato42, line 28: I think we can state explicitly here that re-enabling it will make your application vulnerable if an attacker can force it to process arbitrary ciphertexts.

tomato42 (changes requested):
Test coverage to verify that attempt of creation or processing of RSA1_5 ciphertexts will cause predictable and readable exceptions?

simo5:
Yeah I meant to add one test, like for null, I'll do so and rebase

tomato42:
I think it should be two: one for receiving RSA1_5 and one for sending

also, there's a typo in commit message: `is not contacnt time.`
and shouldn't it be `code needs to be` not `code need to be`?

simo5:
I added a test that is sufficient, algorithms are enable globally, so there is no point in making multiple testes.

Do we have some generic language for the readme to explain what the vulnerability will be, I am coming short on a succinct way to describe the risk.

tomato42:
> I added a test that is sufficient, algorithms are enable globally, so there is no point in making multiple testes.
 
ok

> Do we have some generic language for the readme to explain what the vulnerability will be, I am coming short on a succinct way to describe the risk.

how about:
```
Note: if you enable support for `RSA1_5` and the attacker is able to measure the processing
times in your application of ciphertexts selected by them, then your application will be
vulnerable to a Bleichenbacher RSA padding oracle, allowing the so-called "Million messages
attack", first published in 1998. That attack allows the attacker to decrypt messages it intercepted
(even if they were encrypted with RSA-OAEP) or forge signatures (both RSA-PKCS#1 v1.5 and
RSASSA-PSS).
```
not succinct but should be "scary" (blunt) enough

simo5:
This is perfect, I stole most of it and amended, please review.

tomato42 (approved):
looks good to me

PR #197, opened 2020-12-16
Make PBES2 behave like all other algorithms

### LINKED ISSUE #196: Encryption with PBES2-HS256+A128KW fails to 'JWK' object has no attribute 'encode'
Adjusting the example at https://jwcrypto.readthedocs.io/en/latest/jwe.html#symmetric-keys to use `PBES2-HS256+A128KW` for `alg` leads to error:

```py
...
>>> jwetoken = jwe.JWE(
        payload.encode('utf-8'),
        json_encode({"alg": "PBES2-HS256+A128KW", "enc": "A256CBC-HS512"})
    )
>>> jwetoken.add_recipient(key)
AttributeError: 'JWK' object has no attribute 'encode'
```

Unlike `_AesGcmKw._get_key()` and `_AesKw._get_key()`, `_Pbes2HsAesKw._get_key()` seems to expect raw `str`/`bytes` instead of a `JWK` instance. So this does work:

```py
>>> jwe_token.add_recipient('my-secret')
```

Why is that? Could it documented somewhere, or did I miss something?

simo5:
PBES2 accepts only a plaintext password or bytes which is different
from all other algorithms. Make it support passing in a symmetric key,
and add a jwk classmethod to create a symmetric key from a password.

Resolves #196

### DIFF
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -1052,6 +1052,21 @@ class JWK(dict):
         except KeyError:
             raise AttributeError
 
+    @classmethod
+    def from_password(cls, password):
+        """Creates a symmetric JWK key from a user password.
+
+        :param key: A password in utf8 format.
+        """
+        obj = cls()
+        params = {'kty': 'oct'}
+        try:
+            params['k'] = base64url_encode(password.encode('utf8'))
+        except Exception as e:  # pylint: disable=broad-except
+            raise InvalidJWKValue(e)
+        obj.import_key(**params)
+        return obj
+
 
 class _JWKkeys(set):
 
--- a/jwcrypto/jwa.py
+++ b/jwcrypto/jwa.py
@@ -583,10 +583,15 @@ class _Pbes2HsAesKw(_RawKeyMgmt):
         self.aeskwmap = {128: _A128KW, 192: _A192KW, 256: _A256KW}
 
     def _get_key(self, alg, key, p2s, p2c):
-        if isinstance(key, bytes):
-            plain = key
+        if not isinstance(key, JWK):
+            # backwards compatiblity for old interface
+            if isinstance(key, bytes):
+                plain = key
+            else:
+                plain = key.encode('utf8')
         else:
-            plain = key.encode('utf8')
+            plain = base64url_decode(key.get_op_key())
+
         salt = bytes(self.name.encode('utf8')) + b'\x00' + p2s
 
         if self.hashsize == 256:
--- a/jwcrypto/tests.py
+++ b/jwcrypto/tests.py
@@ -558,6 +558,11 @@ class TestJWK(unittest.TestCase):
         key['unknown'] = 2
         self.assertFalse(key.unknown == key['unknown'])
 
+    def test_jwk_from_password(self):
+        key = jwk.JWK.from_password('test password')
+        self.assertEqual(key['kty'], 'oct')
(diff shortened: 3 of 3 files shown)


jwcrypto/jwk.py
```diff
@@ -1052,6 +1052,22 @@ def __setattr__(self, item, value):
...
+    @classmethod
+    def from_password(cls, password):
+        """Creates a symmetric JWK key from a user password.
+
+        :param key: A password in utf8 format.
+        """
+        obj = cls()
+        params = dict()
+        try:
+            params['kty'] = 'oct'
```
tuukkamustonen, line 1064: You could leave this out of try-except, but doesn't really hurt to have it there as it should never fail anyway

tuukkamustonen:
Looks good as far as I can tell (which is unfortunately very little).

jwcrypto/jwk.py
simo5: good point

simo5:
> Looks good as far as I can tell (which is unfortunately very little).

Does this interface work for you?
Can you test it gives you a compatible JWE to the one you expect ?

tuukkamustonen:
I would say it's good.

I would be constructing JWK directly via constructor or `import_key()`, rather than through `from_password()` because I want to specify also `kid` (for multi-key decryption, if that gets added) and perhaps even `alg` as a hint of key details (I believe that's allowed).

Or I'll just use `from_password()` and forget going hi-fi. It's nice (but maybe it could allow `**kwargs` for any extra fields).

In any case, the addition to `_get_key()`, which allows passing `JWK` instance now, is the real gold nugget here.

I tested it (by patching my currently installed version) and I was able to decrypt the token using `JWK` instance now. So, seems to work 👍🏻.

tuukkamustonen:
> I want to specify also kid (for multi-key decryption, if that gets added)

The JOSE header in our current implementation doesn't seem to contain `kid` at all, but I guess that's just oversight/ignorance of it not being added, and not limitation of the spec.

tuukkamustonen:
It's 0:23 over here in Eastern Europe. I'll check in again tomorrow if you want more test-driving.

simo5:
> I would say it's good.
> 
> I would be constructing JWK directly via constructor or `import_key()`, rather than through `from_password()` because I want to specify also `kid` (for multi-key decryption, if that gets added) and perhaps even `alg` as a hint of key details (I believe that's allowed).

JWK['kid'] = <whatever> :-)
Unfortunately we do not add thr kid from the key to the generated JWE, but you can pass it as a header to add to the JWE explicitly by referencing 'kid': JWK['kid'] when constructing the header dictionary.

> Or I'll just use `from_password()` and forget going hi-fi. It's nice (but maybe it could allow `**kwargs` for any extra fields).

Yes I'll add it so you can pass a kid or any future parameters.

> In any case, the addition to `_get_key()`, which allows passing `JWK` instance now, is the real gold nugget here.
> 
> I tested it (by patching my currently installed version) and I was able to decrypt the token using `JWK` instance now. So, seems to work 👍🏻.

Excellent, thanks!

simo5:
> Yes I'll add it so you can pass a kid or any future parameters.

On second thought, I think I will keep this simple, an additional parameter can easily be added like:
key = jwk.JWK.from_password('foo')['kid'] = 'baz'

and any other can be easily added similarly as well. it is unlikely to be common anyway.

Issue #198, opened 2021-01-09
JWCrypto allows creation of invalid JWS serializations


mrbax:
```
>>> jwk = jwcrypto.jwk.JWK.generate( kty = 'oct' )
>>> jws = jwcrypto.jws.JWS( f'{{"exp": { int( time.time() ) + 60 }}}' )
>>> jws.add_signature( jwk, alg = 'HS256' )
>>> jws.verify( jwk, alg = 'HS256' )
>>> print( jws.serialize( True ) )
.eyJleHAiOiAxNjEwMTgxMjQ1fQ.uiq38Mpw_lM5QaEuimB7Qq0k8Oypk4svDS-MU_ndYQI
```

simo5:
@mrbax do you think JWCrypto should raise if a protected header is not available ?

simo5:
@mrbax or do you think that JWCrypto should syntethize a protected header with the appropriate 'alg' if one is not provided ?

I guess the latter may make sense and I am slightly inclined in that direction.

mrbax:
I vote for synthesizing a protected header if necessary and raising an exception if not possible

simo5:
Looking at existing code and also the JOSE Cookbook (RFC 7520 section 4.7), I think the only solution is to raise if someone requests the compact serialization and there is no protected header.
So I will go in that direction.

simo5:
I think #199 is the best I can do for now ... PTAL

simo5:
Adding back link to the originating issue as it has useful information:
https://github.com/jsonwebtoken/jsonwebtoken.github.io/issues/513

mrbax:
So that's what PTAL means!  +1.  🙂

PR #199, opened 2021-01-11
Enforce protected header in compact serilization

simo5:
As RFC 7515 Sections 7.10 states the compact serialization must carry
the algorithm in the protected header. Make sure we do not produce a
compact serialization without that.

Fixes: #198


simo5:
@mrbax PTAL

simo5:
I think this is the correct thing to do even without formal review. Merging.

PR #200, opened 2021-01-12
leeway fix

BuffaloDan:
Leeway is subtracted from the expiration limit, thus giving less time instead of more (similar with _check_nbf()).
Correct me if I'm wrong, but this does not seem to be the intended behaviour.


BuffaloDan:
I was wrong.

simo5:
No worries, thank you for your contribution anyway!

PR #210, opened 2021-06-03
Fix crash in Keyset JWT decoding exception handler

### LINKED ISSUE #209: crash on absent key id during exception handling
This has been introduced by 0edf66d59b81465dda58600c8dd0c715affbf6ee in 0.9:
```
--- a/jwcrypto/jwt.py
+++ b/jwcrypto/jwt.py
@@ -489,7 +489,7 @@ class JWT(object):
                         self.deserializelog.append("Success")
                         break
                     except Exception as e:  # pylint: disable=broad-except
-                        keyid = k.key_id
+                        keyid = k['kid']
                         if keyid is None:
```
The property key_id returned None if key had no key id, but this made 'kid' mandatory.

The bug happens whenever an exception is raised during deserialization with such key. For example, a call to `JWT(jwt=jwt, key=key, algs=['RS256', 'ES256'])` with an ES256 key will trigger it, as the first attempt to use key as a RS256 key raises an exception.

simo5:
Fixes #209

### DIFF
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -305,7 +305,6 @@ class JWK(dict):
             are provided.
         """
         super(JWK, self).__init__()
-        self._params = dict()
 
         if 'generate' in kwargs:
             self.generate_key(**kwargs)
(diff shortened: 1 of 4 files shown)


simo5:
@Zil0-eo this should fix the issue

Zil0-eo:
It does, thanks :)

Note that there are two places where `.key_id` has been replaced by `['kid']`, the second one is https://github.com/latchset/jwcrypto/commit/0edf66d59b81465dda58600c8dd0c715affbf6ee#diff-c30947f8244c914e06a5c138fd70c7a0759bb70b300c980d2744df399466cf01R1134. Maybe a check that 'kid' is in jwk should be added there too, just to be safe?

simo5:
Interestingly if I change this other occurrence to use get('kid') I get test failures in unexpected places ...

simo5:
Welp ... this is a problem:
`>>> k['kid'] = 'foo'
>>> k['kid']
'foo'
>>> k.get('kid')
>>> 
`

simo5:
I pushed a patch that "fixes" the tests, but I am pretty sure @tiran will scold me for that :-D
I will discuss with Christian this area and find a decent solution tomorrow.

simo5:
Changed code to remove _param use completely, and therefore implementing .get() is now not necessary anymore and I removed that commit.

tiran (approved):

PR #211, opened 2021-06-03
Fix keyset import with similar keys

### LINKED ISSUE #208: New __hash__() method doesn't consider kid
Version 0.9 introduced a new JWK.__hash__() method that does not consider kid.

https://github.com/latchset/jwcrypto/commit/0edf66d59b81465dda58600c8dd0c715affbf6ee

When using JWKSet.import_keyset(), server keys can be unexpectedly filtered out if some keys contain the same underlying key values, but only differ by other metadata, such as kid.  If client tokens generated using the filtered kid key, authorization failure occurs.

simo5:
Keys that are idential except for their key ID were filtered out in set
import. Change __hash__() to include the value of 'kid' as well.

Also add __eq__() function which should always be provided if __hash__()
is implemented.

Fixes #208


simo5:
@rabbittsoup can you check if you think this is sufficient to fix #208 ?

simo5:
@tiran can you review if this use of __hash__ and __eq__ is now correct ?

rabbittsoup:
This is looking good to me, @simo5.  Thanks for working on it so quickly!

I was curious if it could make sense for JWT.deserialize() to attempt all keys in the case that kid is present but not found?  Instead of raising an exception immediately, try all the keys as if kid was not present before raising the exception if none of them work.  I'm not familiar with related specs or expectations that may dictate the behavior, though.

https://github.com/latchset/jwcrypto/blob/81288c78c6395b64c6e31275ffd28173958054df/jwcrypto/jwt.py#L484

simo5:
We change d the behavior in 0.9, previously we were always going through all keys, but this spammed the log (for each failure until right key was found), and was slower because it was going through keys sequentially.
I am wary of changing behavior again, but will if there is a reasoned approach to be taken.

simo5:
We discussed this with @tiran a few days later, and in fact he did propose the actual change.
So I'll assume this as reviewed and agreed, tests confirm the fix (I manually tested w/o the __hash__ fix and the test did break.

PR #225, opened 2021-07-05
Allow empty payloads in JWS tokens

### LINKED ISSUE #224: No support for empty payloads
I just run into the below mentioned issue:
https://community.letsencrypt.org/t/invalid-status-value-error-msg-post-as-get-action/140556/14

JWS.add_signature does not allow empty (e.g. empty string) payloads which is requires for getting challenges via the ACME protocol.
Anybody else ran into this problem? Not sure if fixing is trivial, as I basically like jwcrypto and don't want to migrate to another module.

simo5:
Apparently Let's Encrypt's ACME Server accepts JWT tokens that are JWS
tokens with an empty payload.

Fixes #224

### DIFF
--- a/jwcrypto/jws.py
+++ b/jwcrypto/jws.py
@@ -180,8 +180,7 @@ class JWS(object):
         :param header_registry: Optional additions to the header registry
         """
         self.objects = dict()
-        if payload:
-            self.objects['payload'] = payload
+        self.objects['payload'] = payload
         self.verifylog = None
         self._allowed_algs = None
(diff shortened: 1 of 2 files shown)


simo5:
@bbczeuz can you test this fixes your issue?

adontz:
> 
> 
> @bbczeuz can you test this fixes your issue?

I have the same issue and commenting payload emptiness check in your code fixed it, generated valid (accepted by LetsEncrypt ACME server) signature.

PR #229, opened 2021-08-06
Set python requirement to 3.6+ in setup.py

### LINKED ISSUE #228: Python 3+ version not specified in pypi package
With the 1.0 release and python 2.7 depreciation, the package should have specified that it needs python > 3. I believe that in `setup.py` , this should be added:

` ` `python
    python_requires='>=3',`
` ` `

The 1.0.0 tag should also be dropped from pypi, so that the users stilll struggling with the migration to python 3 can ` pip install` the package without pinning the previous version.

clemux:
This should fix #228 

See: https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,4 +27,5 @@ setup(
         'cryptography >= 2.3',
         'deprecated',
     ],
+    python_requires = '>= 3.6',
 )


simo5:
Thanks!

PR #234, opened 2021-08-14
Use list/dict literals instead

atombrella:
I'm not sure if the object-commit really adds any value. You can delete it, if desired. There's no speed difference, unlike the `{}` and `[]` replacements.

### DIFF
--- a/jwcrypto/jwa.py
+++ b/jwcrypto/jwa.py
@@ -82,7 +82,7 @@ def _decode_int(n):
     return int(hexlify(n), 16)
 
 
-class _RawJWS(object):
+class _RawJWS:
 
     def sign(self, key, payload):
         raise NotImplementedError
@@ -333,7 +333,7 @@ class _None(_RawNone, JWAAlgorithm):
     algorithm_use = 'sig'
 
 
-class _RawKeyMgmt(object):
+class _RawKeyMgmt:
 
     def wrap(self, key, bitsize, cek, headers):
         raise NotImplementedError
@@ -849,7 +849,7 @@ class _EdDsa(_RawJWS, JWAAlgorithm):
         raise NotImplementedError
 
 
-class _RawJWE(object):
+class _RawJWE:
 
     def encrypt(self, k, a, m):
         raise NotImplementedError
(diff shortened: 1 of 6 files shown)


simo5:
Is there any specific reason for these patches ?

atombrella:
> Is there any specific reason for these patches ?

Updating to newer Python 3.6+ syntax. And speed (although comparing stuff at the nanosecond-level may not be relevant to you):

```
In [1]: %timeit dict()
The slowest run took 32.78 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 76.8 ns per loop

In [2]: %timeit {}
The slowest run took 41.75 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 30.8 ns per loop

In [3]: %timeit list()
The slowest run took 39.27 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 67.7 ns per loop

In [4]: %timeit []
The slowest run took 86.93 times longer than the fastest. This could mean that an intermediate result is being cached.
100000000 loops, best of 3: 17.7 ns per loop

In [5]: %%timeit
   ...: o = dict()
   ...: o['valid'] = True
   ...: 
The slowest run took 33.63 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 97.9 ns per loop

In [6]: %%timeit
   ...: o = {'valid': True}
   ...: 
The slowest run took 32.19 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 50.1 ns per loop
```

simo5:
Ok, I do not care much for speed in this context, but I see no harm either.

atombrella:
The failures all appear to be related to `apt-get` not being able to get the `cargo` package.

simo5:
> The failures all appear to be related to `apt-get` not being able to get the `cargo` package.

Yeah I'll try to re-run the tests, seem like an issue with github action caching

simo5:
Looks like there is a transient issue with repositories, I will try to re-run the tests again tomorrow and hopefully it will be fixed by then

simo5:
I should have fixed the apt-get update issue via #236

simo5:
I have now rebased your branch to master and force-pushed it to your repository
If tests pass I will merge the PR

simo5:
All test but CodeQL pssed, codeQL was already passing earlier, so I am just going to push now

PR #266, opened 2022-04-12
Handle unique keys with duplicate kid in JWKSet

jusdino:
I recently ran into a case where token servers had unique keys but with duplicate `kid`s. Looking at the RFC, it has pretty loose requirements for `kid` and doesn't even require that `kid` be unique:

>[4.5](https://datatracker.ietf.org/doc/html/rfc7517#section-4.5).  "kid" (Key ID) Parameter
> 
>    The "kid" (key ID) parameter is used to match a specific key.  This is used, for instance, to choose among a set of keys within a JWK Set during key rollover.  The structure of the "kid" value is unspecified.  When "kid" values are used within a JWK Set, different keys within the JWK Set SHOULD use distinct "kid" values.  (One example in which different keys might use the same "kid" value is if they have different "kty" (key type) values but are considered to be equivalent alternatives by the application using them.)  The "kid" value is a case-sensitive string.  Use of this member is OPTIONAL. When used with JWS or JWE, the "kid" value is used to match a JWS or JWE "kid" Header Parameter value.

I think, the way the JKWSet is written now, it would not properly handle the example case, since it may find the different `kty` key and fail there, rather than moving on to the key with the correct `kty`. So, rather than assume that the first key with a matching `kid` is the only correct key, better to try each key with a matching kid.

### DIFF
--- a/jwcrypto/jws.py
+++ b/jwcrypto/jws.py
@@ -289,14 +289,29 @@ class JWS:
         c = JWSCore(a, key, protected, payload, self._allowed_algs)
         c.verify(signature)
 
-    def verify(self, key, alg=None):
+    # Helper to deal with detached payloads in verification
+    def _get_obj_payload(self, obj, dp):
+        op = obj.get('payload')
+        if dp is not None:
+            if op is None or len(op) == 0:
+                return dp
+            else:
+                raise InvalidJWSOperation('Object Payload present but'
+                                          ' Deatched Payload provided')
+        return op
+
+    def verify(self, key, alg=None, detached_payload=None):
         """Verifies a JWS token.
 
         :param key: The (:class:`jwcrypto.jwk.JWK`) verification key.
         :param alg: The signing algorithm (optional). Usually the algorithm
             is known as it is provided with the JOSE Headers of the token.
+        :param detached_payload: A detached payload to verify the signature
+            against. Only valid for tokens that are not carrying a payload.
 
         :raises InvalidJWSSignature: if the verification fails.
+        :raises InvalidJWSOperation: if a detached_payload is provided but
+                                     an object payload exists
         """
 
         self.verifylog = []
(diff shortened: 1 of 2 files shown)


simo5:
This is interesting ... of course this is accepting that messed up servers are ok, but that's life I guess.

On the patch itself one concern I have is that get_key() remains functional and returns the first match without indication that something may be wrong.
I tink we may want to change get_key() implementation to raise an error if we find more than one key.
Something like:
```
def get_key(self, kid):
   jwks = self['keys'].get_keys(kid)
   if jwks is None:
       return None
   if len(jwks) > 1:
       raise ValueError("Duplicate kids found use get_keys()")
   return jwks[0]
```

simo5:
The other problem I see is duplicated storage in _JWKkeys(), ti is fragile and can lead to inconsistencies if someone tries to remove elements from the set.
Given _JWKkeys is an internal private class I think we should simply change the code to either store in  adifferent way, or iterate on the set every time w/o additional _by_kid storage.

simo5:
I prefer iterating, and unifying the get_key[s] interfaces with a n internal retrieve operation that is given a boolean named something like single = true/false
if true then iteration will always go through the whole set but will raise a if a second kid is found (this is called from get_key()) and just add to the list otherwise (for get_keys())

jusdino:
Those things all make sense to me. I looked at similar options to what you considered but landed where I did because of a few considerations:

- I wanted to avoid breaking changes as best I could so I left `get_keys()` entirely alone, not knowing how clients used that public method.
- I initially added more iteration on the `get_keys()` method but my teammate pointed out that it would be more efficient to add some compute on `.add()` and avoid extra iteration work on validation since the former happened a lot less often than the latter, with the cost that you pointed out of having the data structure inside _JWKkeys be more brittle - I didn't consider _removing_ keys though.

I think those both are just personal preference though. I can update to the approach you've identified. Here shortly.

simo5:
Looks good to me, please rebase into a single commit.

simo5:
Ah and please fix the pep8 issue (line longer than 79 characters)

PR #268, opened 2022-04-13
handle keys with duplicate kids

jusdino:
Well, crud. I accidentally close #266 it seems. Here is the description, for reference.

I recently ran into a case where token servers had unique keys but with duplicate `kid`s. Looking at the RFC, it has pretty loose requirements for `kid` and doesn't even require that `kid` be unique:

>[4.5](https://datatracker.ietf.org/doc/html/rfc7517#section-4.5).  "kid" (Key ID) Parameter
> 
>    The "kid" (key ID) parameter is used to match a specific key.  This is used, for instance, to choose among a set of keys within a JWK Set during key rollover.  The structure of the "kid" value is unspecified.  When "kid" values are used within a JWK Set, different keys within the JWK Set SHOULD use distinct "kid" values.  (One example in which different keys might use the same "kid" value is if they have different "kty" (key type) values but are considered to be equivalent alternatives by the application using them.)  The "kid" value is a case-sensitive string.  Use of this member is OPTIONAL. When used with JWS or JWE, the "kid" value is used to match a JWS or JWE "kid" Header Parameter value.

I think, the way the JKWSet is written now, it would not properly handle the example case, since it may find the different `kty` key and fail there, rather than moving on to the key with the correct `kty`. So, rather than assume that the first key with a matching `kid` is the only correct key, better to try each key with a matching kid.

### DIFF
--- a/jwcrypto/tests.py
+++ b/jwcrypto/tests.py
@@ -437,6 +437,24 @@ class TestJWK(unittest.TestCase):
             num += 1
         self.assertEqual(num, len(PrivateKeys['keys']))
 
+    def test_jwkset_get_keys(self):
+        # Test key set with mutiple keys
+        ksm = jwk.JWKSet.from_json(json_encode(PrivateKeys))
+        k1 = jwk.JWK.from_json(json_encode(PrivateKeys['keys'][0]))
+        kwargs = RSAPrivateKey.copy()
+        kwargs['kid'] = '1'
+        k2 = jwk.JWK(**kwargs)
+        self.assertEqual(k1, ksm.get_key('1'))
+        self.assertIsNone(ksm.get_key('not-there'))
+
+        ksm.add(k2)
+        self.assertEqual({k1, k2}, ksm.get_keys('1'))
+        self.assertEqual(3, len(ksm['keys']))
+        # Expect that duplicate kids will
+        # raise an exception when we use get_key
+        with self.assertRaises(jwk.InvalidJWKValue):
+            ksm.get_key('1')
+
     def test_jwkset_issue_208(self):
         ks = jwk.JWKSet()
         key1 = RSAPrivateKey.copy()
@@ -1482,7 +1500,7 @@ class TestJWT(unittest.TestCase):
         t = jwt.JWT(header, A1_claims, algs=jwe_algs_and_rsa1_5)
         t.make_encrypted_token(key)
         token = t.serialize()
-        self.assertRaises(jwe.InvalidJWEData, jwt.JWT, jwt=token, key=keyset)
+        self.assertRaises(jwt.JWTMissingKey, jwt.JWT, jwt=token, key=keyset)
 
         keyset = jwk.JWKSet.from_json(json_encode(PrivateKeys))
         # encrypt a new JWT with no kid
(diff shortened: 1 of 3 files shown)


simo5:
Ah pep8 still fails.

(I can fix commit issues for you if you run into trouble) just make sure to "allow changes by the repository mainatainers"

jusdino:
Not sure why, but I don't seem to have the 'allow changes by repository maintainers' check box :shrug:

jwcrypto/tests.py
```diff
@@ -437,6 +437,30 @@ def test_jwkset(self):
...
+        # Test key set with mutiple keys
+        ksm = jwk.JWKSet.from_json(json_encode(PrivateKeys))
+        k1 = jwk.JWK.from_json(json_encode(PrivateKeys['keys'][0]))
+        kwargs = RSAPrivateKey.copy()
+        kwargs['kid'] = '1'
+        k2 = jwk.JWK(**kwargs)
+        num = 0
+        for item in ksm:
+            self.assertTrue(isinstance(item, jwk.JWK))
+            self.assertTrue(item in ksm)
```
simo5, line 450: a bit redundant, you are iterating over ksm in which case this can fail?

jwcrypto/tests.py
```diff
@@ -1497,6 +1521,29 @@ def test_decrypt_keyset(self):
...
+
+        # encrypt a new JWT with kid
+        header = copy.copy(A1_header)
+        header['kid'] = 'testkey'
+        t = jwt.JWT(header, A1_claims, algs=jwe_algs_and_rsa1_5)
+        t.make_encrypted_token(key)
+        token = t.serialize()
+
+        # try to decrypt without a matching key
+        self.assertRaises(jwt.JWTMissingKey, jwt.JWT, jwt=token, key=keyset)
```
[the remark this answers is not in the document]
simo5: As far as I can tell this test is only testing that algs parameter is missing, can probably simply be removed...
which probably means I have a bug in the test you copied from :-/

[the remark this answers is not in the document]
jusdino: That was a mistake. Pushing test fix.

jwcrypto/tests.py
jusdino: Good point. Removed.

jwcrypto/tests.py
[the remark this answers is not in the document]
jusdino: FWIW, I copied that from `test_decrypt_keyset()` if you want to check that is doing what you intended.

jwcrypto/tests.py
```diff
@@ -1497,6 +1515,31 @@ def test_decrypt_keyset(self):
...
+        key = jwk.JWK(kid='testkey', **E_A2_key)
+        header = copy.copy(A1_header)
+        header['kid'] = 'testkey'
+        t = jwt.JWT(header, A1_claims, algs=jwe_algs_and_rsa1_5)
+        t.make_encrypted_token(key)
+        token = t.serialize()
+
+        # try to decrypt without a matching key
+        with self.assertRaises(jwt.JWTMissingKey):
+            jwt.JWT(jwt=token, key=keyset, check_claims={'exp': 1300819380})
```
simo5, line 1534: you need to add algs=jwe_algs_and_rsa1_5 here or you are not testing the "right" thing
somehow, still trying to figure out why it was passing previously

jwcrypto/tests.py
```diff
@@ -1497,6 +1515,31 @@ def test_decrypt_keyset(self):
...
+
+        # add right key
+        keyset.add(key)
+
```
simo5, line 1538: This will definitely now break the way it should, I may just merge and fix the test myself later for the false nergative I've seen

simo5 (approved):
As long as CI is happy I will be ok with this now and then fix the false negative in a follow up myself

simo5:
Tests are happy, I am going to merge soon and then deal with the false negative myself.

Thanks for your contribution!

PR #290, opened 2022-05-13
Try to disable HW optimizations on ppc64le

simo5:
As builds started failing with latest Fedora/openssl try to see if it is
due to HW optimizations failing on the github runner.


simo5:
This seem to have fixed issues in a test run.
Given ppc64le runs take 35/40 minutes I am just going to merge now, and handle again later should this not be sufficient to restore ppc64le to working conditions.

PR #292, opened 2022-05-20
Introduce a new JWKeyNotFound exception

### LINKED ISSUE #291: 1.3.0 breaks compatibility and should have been 2.0.0
https://github.com/latchset/jwcrypto/commit/659fe723d1be917c45007dcd6efab40e3e0b7a49 is a compatibility breaking change in this library's API, as the `JWTMissingKeyID` exception is no longer thrown.  This should have been associated with a move to version 2.0.0, not 1.3.0.

Is there any chance we can roll back on this change, or add a similar feature to the version in JWS/JWE? It's unclear to me how a user is supposed to track when a signature error is due to a missing key or when it is from the myriad of other potential sources without that exception. This will lead to unnecessary work for every user of JWKSet, as they have to treat all `InvalidJWSSignature` as missing keys. Thus if they are requesting public keys from an external server (the common use case for JWTs), they will need to make that request every time.

simo5:
This new Exception is returned only for the newly introduced support for
using JWKset.
This patch also includes a bugfix for jwe to be able to successfully
decrypt using a JWKSet, which was non-functional, and a direct test for
both JWE and JWS to insure no regressions in JWKSet support.

Also restores use of JWTMissingKey for backwards compatibility.

Fixes #291

### DIFF
--- a/docs/source/common.rst
+++ b/docs/source/common.rst
@@ -36,3 +36,6 @@ Exceptions
 
 .. autoclass:: jwcrypto.common.InvalidJWSERegOperation
    :show-inheritance:
+
+.. autoclass:: jwcrypto.common.JWKeyNotFound
+   :show-inheritance:
--- a/jwcrypto/common.py
+++ b/jwcrypto/common.py
@@ -126,6 +126,22 @@ class InvalidJWSERegOperation(JWException):
         super(InvalidJWSERegOperation, self).__init__(msg)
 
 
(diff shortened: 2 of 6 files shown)


simo5:
@nonathaj this should fix #291 can you confirm it works for you?

bturk-hirez:
@simo5, I work with @nonathaj and can confirm this works for our use case. Thanks for the quick turn around!

PR #301, opened 2022-09-14
Add more heuristics for backwards JWT compat

simo5:
These additional heuristics help in case the calling application was correctly calssifying key usage, as this is another valid hitn of what the application intended.
Invalid key usage would already cause failure, so this does not affect the countermeasures introduced but can avoid issues in older applications.


simo5:
This has been independently tested to resolve issues in FreeIPA where the change caused a failure.

PR #303, opened 2022-09-15
Fix typo in new backwards JWT compat heuristics

jcgruenhage:
I just noticed that the test suite of a package I'm working on are suddenly failing. Short digging later, there was a typo in the newly released 1.4.1


simo5:
Argh, and of course this is one of the heuristics I forogt to add a test for ...

simo5:
Would you mind adding a test in the tests file which checks this condition?
If not I can do it.

I will release 1.4.2 right after we get this in, sigh.

simo5:
NVM, I could not resist and pushed a commit on your branch, so I can deal with this quickly

simo5:
Thanks for spotting this, very appreciated

PR #307, opened 2022-10-04
Fix typos with codespell

cclauss:
[`codespell --ignore-words-list="ect,ser" -w .`](https://pypi.org/project/codespell/)


simo5:
Ah thanks, good catches!
Is codespell available in common linux distributions?
I am thinking we may want to add it as a check so that we catch these mispellings at PR time in the future.

cclauss:
`codespell` is pure Python so `python3 -m pip install codespell` will work.
I install it with `pipx install codespell`.
It also work well in `pre-commit`.

PR #324, opened 2023-05-30
Better support for algorithms that have different input keysize requirement

### LINKED ISSUE #302: Key size on generation is not always computed correctly
For algorithms like A256CBC-HS512 the output keysize is 256 bits, but the input keysize is expected to be 512 bits.

When a JWK(generate='oct', alg='A256CBC-HS512') command is invoked the code tries to determine the default key size to be used, and it does that by inquiring the JWA algorithm for its keysize.

Unfortunately the reported key size for these algorithms is the output key size and not the input keysize ...

simo5:
Fixes #302


raildo (approved):
LGTM

simo5:
Thanks @raildo !

PR #325, opened 2023-06-27
Timezone-aware datetime Object Support

Flat-Irons:
Allows for datetime objects that are timezone-aware to be used for time-based claims: 'exp' and 'nbf' when making a JWT object, converting them to integer unix timestamps.

### DIFF
--- a/jwcrypto/jwt.py
+++ b/jwcrypto/jwt.py
@@ -3,6 +3,7 @@
 import copy
 import time
 import uuid
+from datetime import datetime
 
 from deprecated import deprecated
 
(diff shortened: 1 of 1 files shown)


jwcrypto/jwt.py
```diff
@@ -225,6 +228,14 @@ def __init__(self, header=None, claims=None, jwt=None, key=None,
             self._check_claims = check_claims
 
         if claims is not None:
+            # Check for datetime objects
+            if 'exp' in claims and isinstance(claims['exp'], datetime):
+                # Check if timezone aware
+                claims['exp'] = self._check_and_convert_dt(claims['exp'], 'exp')
```
simo5, line 234: I do not think we should modify the array being passed in.
We should instead make a copy in self.claims and modify them there

jwcrypto/jwt.py
```diff
@@ -388,6 +399,18 @@ def expected_type(self, v):
...
+    def _check_and_convert_dt(self, dt, claim_prop):
+        if (
+            dt.tzinfo is not None
+            and dt.tzinfo.utcoffset(dt) is not None
+        ):
+            dt_timestamp = int(dt.timestamp())
+            return dt_timestamp
+        else:
+            raise ValueError(
+                f"'{claim_prop}' datetime object must be timezone aware"
```
simo5, line 411: Why does it need to be timezone-aware? Can't we just assume UTC if no timezone is set?

simo5 (changes requested):
Nits, 
also please cleanup the commits, needs sign-off line.
Also please squash into a single comment w/o merge stuff and provide a commit message.

simo5:
Please add a test that verifies this si working properly.

Issue #339, opened 2024-02-07
Drop Python 3.6 support since 1.5.2


bobleckelibm:
Per comments in https://github.com/latchset/jwcrypto/pull/337#issuecomment-1932694930 Python 3.6 has been EOL for a long time now — it would make sense to remove official support for it in this package, especially since v1.5.2 where code was introduced that was no longer 3.6 compatible.

Suggested change:
https://github.com/latchset/jwcrypto/blob/main/setup.py#L43

I suggest you go to 3.8 since that still has security support until later this year so why make it too restrictive. But wouldn't be mad if it just bumped to 3.9 or straight to 3.11

bobleckelibm:
I would appreciate if the 1.5.2 on PyPi was yanked and re-released so that 3.6 installs don't pull in the problematic code. But I understand if that is not possible

simo5:
The only thing I can do is to make a 1.5.3 release, which is cheap, so I guess I'll just do that..

bobleckelibm:
Thanks for the amazingly fast response. Given the release went out, the effects aren't all that useful for us unless 1.5.2 is removed because we'll still get it when we try to upgrade jwcrpyto in a 3.6 env. I'm wondering if that's at all possible to do? We're going ahead and pinning for 3.6 anyhow, just want to check

simo5:
Yanked 1.5.2 from pypi

bobleckelibm:
Super thanks again

PR #340, opened 2024-02-07
Drop python 3.6 and 3.7 and add 3.11 support

simo5:
Related to #339


simo5:
@bobleckelibm can you tell me if this looks reasonable ?

bobleckelibm (approved):
Looks great — thank you @simo5 for the quick response

PR #351, opened 2024-04-04
JWE: allow general (non flattened) serialization syntax

Thomas-Mollard:
When we call `.serialize()` on a JWE with 1 single recipient, it uses the [flattened serialization syntax](https://datatracker.ietf.org/doc/html/rfc7516#section-7.2.2) where the member of the unique recipient (`header` and `encrypted_key`) are added in the top-level JSON object instead of in the `recipients`

However, according to the [RFC for JWE](https://datatracker.ietf.org/doc/html/rfc7516#section-7.2.1), the general serialization syntax allows for the JSON to have a `recipients` with only one item

In this PR, I am adding the possibility to use the general (non flattened) serialization syntax by setting `flattened=False` when initializing the `JWE` (`flattened=True` is set by default)

The change I did is the "easiest" one I see as it doesn't cause any breaking change  and only updates the internal `objects` when adding a recipient as is expected depending on the general/flattened serialization used

We could also:
- keep `flattened` in `__init__`, always add recipient in `recipients` when calling `.add_recipient` and serialize differently depending on `self.flattened` when calling `.serialize`
- always add recipient in `recipients` when calling `.add_recipient`, add a bool flag `flattened` on `serialize` without the check on number of `recipients` (different as `compact`) and serialize differently depending on the flag`flattened` when calling `.serialize`

Let me know if you prefer than I implement this in another way

I'm not sure where I need to put the test, I would guess in `TestJWE` and add a `test_general_serialization`, would that make sense?


jwcrypto/jwe.py
```diff
@@ -251,19 +253,26 @@ def add_recipient(self, key, header=None):
...
-            self.objects['recipients'] = []
-            n = {}
-            if 'encrypted_key' in self.objects:
-                n['encrypted_key'] = self.objects.pop('encrypted_key')
-            if 'header' in self.objects:
-                n['header'] = self.objects.pop('header')
-            self.objects['recipients'].append(n)
-            self.objects['recipients'].append(rec)
+        if self.flattened:
+            if 'recipients' in self.objects:
```
simo5, line 257: pull the if 'recipients' part out, as it is always the same:
```
if 'recipients' in self.objects:
    self.objects['recipients'].append(rec)
elif self.flattened:
    if 'encrypted_key' in self.objects or 'header' in self.objects:
    [...]
else:
    self.objects['recipients'] = [rec]
```

simo5 (changes requested):
Looks good, but please refactor the if/elif/else as described.

Also add at least 1 test where you set se flattened to false and then verify that you get an object with only one recipient in it.

jwcrypto/jwe.py
```diff
@@ -253,17 +255,21 @@ def add_recipient(self, key, header=None):
...
+                if 'header' in self.objects:
+                    n['header'] = self.objects.pop('header')
+                self.objects['recipients'].append(n)
+                self.objects['recipients'].append(rec)
+            else:
+                self.objects.update(rec)
         else:
-            self.objects.update(rec)
+            self.objects['recipients'] = [rec]
+
```
simo5, line 272: remove this line to make pep8 happy and I'll merge.

simo5 (changes requested):
LGTM, just a nit to make syntax tests happy

jwcrypto/jwe.py
simo5: ah you just did push a new commit for this :)

Thomas-Mollard: Yes I missed this the first time, sorry for the back and forth!

simo5:
please squash everything in a single commit with a comment that briefly explain what this change is about

simo5 (changes requested):
LGTM, just need to squash all in a single commit

Thomas-Mollard:
> LGTM, just need to squash all in a single commit

Thanks!
Is it OK now?

simo5:
> > LGTM, just need to squash all in a single commit
> 
> Thanks! Is it OK now?

A little more meat on what the change did would have been nice, but it is good enough, I will merge when tests are all green

simo5 (approved):

Thomas-Mollard:
Thanks a lot for your help and handling this so quickly!
Do you plan to make a new release including this change soon by any chance?

simo5:
Making a release is not hard work, if you have a need we can arrange to make one soon.

I generally do not rush them out an instead accumulate a few changes to reduce the churn though. The only exception is security releases, I churn them out as fast as needed.

Thomas-Mollard:
There is no real urgency per se on our side to have a release as we have a workaround but it's not super pretty and it would ultimately be better to use this change

PR #359, opened 2024-07-08
Add support for 'scope' claim with multiple scopes

tony2001:
Claim 'scope' is formally standardized by this RFC:
https://datatracker.ietf.org/doc/html/rfc8693#name-scope-scopes-claim

It's been in use for quite a while though and it's a part of standard token provided by Keycloak, which is a long time industry standard for access management.
Keycloak tokens may and quite often do contain multiple scopes, hence the need for this patch.

### DIFF
--- a/jwcrypto/jwt.py
+++ b/jwcrypto/jwt.py
@@ -479,6 +479,7 @@ class JWT:
         self._check_string_claim('iss', check_claims)
         self._check_string_claim('sub', check_claims)
         self._check_array_or_string_claim('aud', check_claims)
+        self._check_string_claim('scope', check_claims)
         self._check_integer_claim('exp', check_claims)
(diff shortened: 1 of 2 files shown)


simo5:
Code looks good thanks!
Please fix the style issues, you can test them locally using tox

simo5 (changes requested):
Style, otherwise LGTM

tony2001:
I'm not really familiar with Python, but I think I've fixed the issues.
Could check again pls?

simo5 (approved):
LGTM

simo5:
The failing test is codespell being stupid and flagging a function name in code you have not touched, I'll fix it later.

Thanks for the contribution, really appreciated!

PR #361, opened 2024-12-02
Add functions to generate a similar key

jschlyter:
A common use-case when implementing rekeying or testing is to be able to generate a similar key to an already existing one. This PR adds a `JWK.generate_similar(key)` classmethod to support this, as well as a `JWK.get_generate_params()` to get the parameters used to generate a key. Test cases included.

Perhaps there's an easier way to do this; if so please enlighten me.


simo5:
Looks like a reasonable approach

simo5:
Please squash the fix in the proper patches and force push the result

PR #375, opened 2026-04-08
Fix list iteration in claim format validation

simo5:
Update the validation logic to iterate over the items in claims[name] rather than the claims dictionary itself. This fixes a bug where list-based claims (such as multiple audiences) were not correctly checking if all elements were strings. A test for 'aud' claim type validation has been added to prevent future regressions.

Credit to @0xHunSec for reporting this although it was rejected as a security issue, because this flaw gives no security bypass abilities, it is just a bug.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,7 @@
                         "python-311",
                         "python-312",
                         "python-313",
+                        "python-314",
                         "doc",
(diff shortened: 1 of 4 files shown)


rjeffman (approved):
LTGM.

rjeffman (approved):
Changes looks good.

PR #379, opened 2026-05-05
Wrap JWKSet parsing errors in InvalidJWKValue

### LINKED ISSUE #378: insufficient validation in `JWKSet.from_json` (`import_keyset`)
One would expect to get an `InvalidJWKValue` value from loading an invalid JWKSet. This is not the case...

```py
import traceback
from jwcrypto.common import JWException
from jwcrypto.jwk import JWKSet

values = [
    'foobar',
    '"keys"',
    '["keys"]',
    '{}',
    '{"keys": {}}',
    '{"keys": {{}}}',
    '{"keys": {"foo": "bar"}}',
    '{"keys": [{"foo": "bar"}]}',
    '{"keys": []}',
    '{"keys": [[]]}',
]

for value in values:
    print(f"'{value}' - ", end="")
    try:
        JWKSet.from_json(value)
        print("valid")
    except JWException as e:
        print(f"valid")
    except Exception as e:
        print(f"invalid")
        print(traceback.format_exc())
```

The above gives us a selection of `TypeError` and `AttributeError` from `import_keyset`.
https://github.com/latchset/jwcrypto/blob/db03d4c37b6020a136bc25373656cce49a700acf/jwcrypto/jwk.py#L1357-L1370

simo5:
Moved the dictionary iteration and key creation logic inside the try-except block. This ensures that any exceptions raised during the instantiation of individual JWK objects or validation checks are properly caught and safely re- raised as an InvalidJWKValue exception, rather than leaking unhandled errors.

Fixes: #378

### DIFF
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -1356,18 +1356,17 @@ class JWKSet(dict):
         """
         try:
             jwkset = json_decode(keyset)
+            if 'keys' not in jwkset:
+                raise ValueError("'keys' not in set")
+
+            for k, v in jwkset.items():
+                if k == 'keys':
+                    for jwk in v:
+                        self['keys'].add(JWK(**jwk))
+                else:
+                    self[k] = v
         except Exception as e:  # pylint: disable=broad-except
             raise InvalidJWKValue from e
-
-        if 'keys' not in jwkset:
-            raise InvalidJWKValue
-
-        for k, v in jwkset.items():
-            if k == 'keys':
-                for jwk in v:
-                    self['keys'].add(JWK(**jwk))
-            else:
-                self[k] = v
 
     @classmethod
     def from_json(cls, keyset):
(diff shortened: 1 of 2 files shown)


rjeffman:
LGTM. I'll give some time for reporter to take a look at this before merging.

rjeffman (approved):

jwcrypto/jwk.py
```diff
@@ -1356,19 +1356,18 @@ def import_keyset(self, keyset):
...
             jwkset = json_decode(keyset)
+            if 'keys' not in jwkset:
+                raise ValueError("'keys' not in set")
+
+            for k, v in jwkset.items():
+                if k == 'keys':
+                    for jwk in v:
+                        self['keys'].add(JWK(**jwk))
+                else:
+                    self[k] = v
```
huwcbjones, line 1367: If we're changing this around, can't we also improve it?
Something like:
```suggestion
            self["keys"].update(JWK(**jwk) for jwk in jwkset.pop("keys"))
            self.update(jwkset)
```

We don't need to raise a `ValueError`, because we'll get a `KeyError` in `pop("keys")` that'll be caught in the except block and re-raised.

simo5: I prefer to keep things readable, and I can't drop the 'keys' check as then I would miss the case when jwkset is an empty dict, which is also invalid.

PR #381, opened 2026-06-05
jwt: add opt-in strict_serialization to enforce compact form

### LINKED ISSUE #342: Jwt.JWT allows parsing tokens with json serialization
According to RFC 7519 only the compact serialization can be used for a proper JWT token.
Jwcrypto accepts any JWS/JWE token for parsing.

While this not a problem per se, people may want to reject even the ability to pars a non compact serialization to reduce potential attack surface.

Provide a "strict" or "serialization" option when instantiating a JWT so that this aspect can be controlled.

arpitjain099:
Fixes #342

RFC 7519 (Section 7.1, step 8 and the definition of a JWT) requires that a JWT be represented using the JWS or JWE Compact Serialization. `JWT.deserialize` today decides what to do based on `jwt.count('.')`: two dots route to a `JWS`, four dots to a `JWE`. Because the JWS/JWE `deserialize` methods try `json_decode` first, a JSON-serialized token whose surrounding JSON happens to contain exactly two (JWS) or four (JWE) `.` characters is accepted as a JWT. That is reachable in practice: a dotted value in an unprotected header (`{"kid":"a.b.c"}`) or a per-recipient header lands the count on 2 or 4, so a JSON Serialization gets parsed where only a compact token should be.

As the issue notes this is not a vulnerability on its own, but parsing a representation the spec disallows is extra attack surface that some callers would rather not expose.

## What this changes

A new opt-in `strict_serialization` keyword argument on `JWT` (default `False`). When it is `True`, `deserialize` rejects any JSON-serialized JWS/JWE with a clear `ValueError` and only accepts the compact form. Detection reuses the same `json_decode` check that `JWS.deserialize` / `JWE.deserialize` already use to tell the two representations apart (a complete compact token never decodes to a JSON object).

The default is `False`, so existing behavior is completely unchanged. This mirrors the existing knob style on this class (`expected_type`, the `JWT_expect_type` module flag) and has no effect on token creation, which already only emits the compact serialization.

## Before / after

```python
# A JSON-serialized JWS whose JSON contains exactly two dots
signer = jws.JWS(payload='{"sub":"alice"}')
signer.add_signature(key, alg='HS256', protected='{"alg":"HS256"}',
                     header={'kid': 'a.b.c'})
json_token = signer.serialize(compact=False)

# Default (unchanged): accepted
jwt.JWT().deserialize(json_token, key)            # ok

# Opt-in strict: rejected
jwt.JWT(strict_serialization=True).deserialize(json_token, key)
# ValueError: Only the JWS/JWE Compact Serialization is allowed for JWTs
#             (RFC 7519), but a JSON Serialization was provided

# Compact token still works in both modes
jwt.JWT(strict_serialization=True).deserialize(compact_token, key)  # ok
```

## Tests

Added `test_jwt_strict_serialization_jws` and `test_jwt_strict_serialization_jwe` covering, for both signed and encrypted tokens:
- flag on: JSON serialization rejected (via `deserialize` and via the constructor `jwt=` path), compact accepted
- flag off (default): JSON serialization still accepted, so the change is backward compatible

Full suite is green (`python -m pytest jwcrypto/tests.py`), flake8 clean on the changed files.

I went with an opt-in flag deliberately to avoid changing the default. If you would rather have a differently named argument (for example `serialization=` taking `"compact"` / `"any"`), or want this surfaced as a module-level switch instead, happy to adjust.

### DIFF
--- a/jwcrypto/jwt.py
+++ b/jwcrypto/jwt.py
@@ -168,7 +168,7 @@ class JWT:
 
     def __init__(self, header=None, claims=None, jwt=None, key=None,
                  algs=None, default_claims=None, check_claims=None,
-                 expected_type=None):
+                 expected_type=None, strict_serialization=False):
         """Creates a JWT object.
 
         :param header: A dict or a JSON string with the JWT Header data.
@@ -190,6 +190,13 @@ class JWT:
          If left to None the code will try to detect what the expected
          type is based on other parameters like 'algs' and will default
          to JWS if no hints are found. It has no effect on token creation.
+        :param strict_serialization: An optional boolean (default False).
+         RFC 7519 mandates that a JWT uses the JWS/JWE Compact
+         Serialization. When set to True any token presented in the
+         JWS/JWE JSON Serialization is rejected at deserialization time,
+         so only the compact representation is accepted. The default is
+         False to preserve backwards compatibility. It has no effect on
+         token creation, which always produces the compact serialization.
 
         Note: either the header,claims or jwt,key parameters should be
         provided as a deserialization operation (which occurs if the jwt
@@ -212,6 +219,7 @@ class JWT:
         self._validity = 600  # 10 minutes validity (up to 11 with leeway)
         self.deserializelog = None
         self._expected_type = expected_type
+        self._strict_serialization = strict_serialization
 
         if header:
             self.header = header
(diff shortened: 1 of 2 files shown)


simo5 (approved):
Excellent proposal, I really like how you addressed it.

simo5:
Thank you!

PR #384, opened 2026-07-20
Add ML-DSA post-quantum signature support (RFC 9964)

rjeffman:
Implement the AKP (Algorithm Key Pair) key type and ML-DSA-44, ML-DSA-65, ML-DSA-87 signing algorithms per IETF RFC 9964. Requires python-cryptography >= 49.0.0.

As python-cryptography 49+ requires Python 3.9+, the support for Python 3.8 in dropped, and tests for Python 3.12, 3,13 and 3,14 were added.

### DIFF
--- a/jwcrypto/jwk.py
+++ b/jwcrypto/jwk.py
@@ -86,11 +86,61 @@ _OKP_CURVES_TABLE = {
 }
 
 
+class UnimplementedAKPAlgorithm:
+    @classmethod
+    def generate(cls):
+        raise NotImplementedError
+
+    @classmethod
+    def from_public_bytes(cls, *args):
+        raise NotImplementedError
+
+    @classmethod
+    def from_seed_bytes(cls, *args):
+        raise NotImplementedError
+
+
+ImplementedAKPAlgorithms = []
+
+try:
+    from cryptography.hazmat.primitives.asymmetric.mldsa import (
+        MLDSA44PublicKey, MLDSA44PrivateKey
+    )
+    ImplementedAKPAlgorithms.append('ML-DSA-44')
+except ImportError:
+    MLDSA44PublicKey = UnimplementedAKPAlgorithm
+    MLDSA44PrivateKey = UnimplementedAKPAlgorithm
+try:
+    from cryptography.hazmat.primitives.asymmetric.mldsa import (
+        MLDSA65PublicKey, MLDSA65PrivateKey
+    )
+    ImplementedAKPAlgorithms.append('ML-DSA-65')
+except ImportError:
+    MLDSA65PublicKey = UnimplementedAKPAlgorithm
+    MLDSA65PrivateKey = UnimplementedAKPAlgorithm
+try:
+    from cryptography.hazmat.primitives.asymmetric.mldsa import (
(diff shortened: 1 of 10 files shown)


jwcrypto/jwk.py
```diff
@@ -86,11 +86,61 @@ def from_private_bytes(cls, *args):
 }
 
 
+class UnimplementedMLDSAKey:
```
simo5: although for now we seem to have only ML-DAS this (and the next) should probably be named something like [Un|I]mplementedAKPAlgorithm[s]

jwcrypto/jwk.py
```diff
@@ -86,11 +86,61 @@ def from_private_bytes(cls, *args):
...
+    from cryptography.hazmat.primitives.asymmetric.mldsa import (
+        MLDSA87PublicKey, MLDSA87PrivateKey
+    )
+    ImplementedMLDSAAlgorithms.append('ML-DSA-87')
+except ImportError:
+    MLDSA87PublicKey = UnimplementedMLDSAKey
+    MLDSA87PrivateKey = UnimplementedMLDSAKey
+
+_MLDSAAlg = namedtuple('MLDSAAlg', 'pubkey privkey')
+_MLDSA_ALG_TABLE = {
```
simo5: this should probably be AKP_ALG_TABLE, even though currently we only have MLDSA algorithms in there.

simo5 (changes requested):
Just some minor nitpicks but otherwise looks good overall.

Are there any test vectors in RFC 9964?
If so I would like to have them in the test suite, if not, we should find some to verify that the code does the right thing.

jwcrypto/jwk.py
rjeffman: Fixed.

jwcrypto/jwk.py
rjeffman: Fixed

rjeffman:
Due to the size of the test vectors, I moved all ML-DSA related tests to `jwcrypto/tests_mldsa.py`.

jwcrypto/jwk.py
```diff
@@ -550,6 +604,42 @@ def _import_pyca_pub_okp(self, key, **params):
...
+                'Algorithm "%s" is not available' % alg)
+        privkey_class = _AKP_ALG_TABLE[alg].privkey
+        key = privkey_class.generate()
+        self._import_pyca_pri_akp(key, alg, **params)
+
+    def _akp_alg_from_pyca_key(self, key):
+        for name, val in _AKP_ALG_TABLE.items():
+            if isinstance(key, (val.pubkey, val.privkey)):
+                return name
+        raise InvalidJWKValue('Invalid ML-DSA Key object %r' % key)
```
simo5: This one should probably say 'Invalid AKP Key Object' now

simo5 (approved):
Just a nit, otherwise this looks good to me.

jwcrypto/jwk.py
rjeffman: Agree. Fixed.

simo5:
Thanks!

PR #389, opened 2026-08-07
Support fully specified algorithms per RFC 9864

### LINKED ISSUE #386: Implement support for Fully specified algorituhms (RFC9864)
https://datatracker.ietf.org/doc/rfc9864/

rjeffman:
Support for fully specified algorithms is added and deprecated EdDSA algorithms are removed from the allowed algorithms list.

As EdDSA is only deprecated, it is not removed and can be forced back if required (see `test_EdDsa_signing_and_verification` test).

Fixes #386

### DIFF
--- a/jwcrypto/jwa.py
+++ b/jwcrypto/jwa.py
@@ -2,6 +2,7 @@
 
 import os
 import struct
+import warnings
 from abc import ABCMeta, abstractmethod
 from binascii import hexlify, unhexlify
 
@@ -41,6 +42,15 @@ This is a security measure to prevent denial-of-service attacks by malicious
 actors providing a very high iteration count.
 """
 
+default_warn_deprecated_algorithms = False
+"""When True, emit a DeprecationWarning when instantiating an algorithm
+whose status is 'deprecated'.
+
+Disabled by default so that consumers are not disrupted by warnings for
+algorithms that remain in wide use (e.g. EdDSA). Set to True to audit
+your code for deprecated algorithm usage.
+"""
+
 default_enforce_hmac_key_length = True
 """Enforces that the HMAC key length is at least the size of the hash
 function's output, as recommended by RFC 7518.
@@ -76,6 +86,12 @@ class JWAAlgorithm(metaclass=ABCMeta):
     @abstractmethod
     def algorithm_use(self):
         """One of 'sig', 'kex', 'enc'"""
+
+    status = 'active'
+    """Algorithm status: 'active', 'deprecated', or 'prohibited'"""
+
+    deprecated_by = None
+    """Replacement algorithm(s) for deprecated algorithms"""
 
     @property
     def input_keysize(self):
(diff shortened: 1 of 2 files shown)


jwcrypto/jws.py
```diff
@@ -30,7 +30,7 @@
     'RS256', 'RS384', 'RS512',
     'ES256', 'ES384', 'ES512',
     'PS256', 'PS384', 'PS512',
-    'EdDSA', 'ES256K', 'Ed25519',
```
simo5, line 33: don't remove EdDSA, even if it is deprecated in theory by the new RFC it will be years before we can actually stop offering it

jwcrypto/tests.py
```diff
@@ -714,13 +714,13 @@ def test_pem_okp(self):
         self.assertTrue(prikey.has_private)
         self.assertTrue(prikey.has_public)
         s = jws.JWS(payload)
-        s.add_signature(prikey, None, {'alg': 'EdDSA'}, None)
+        s.add_signature(prikey, None, {'alg': 'Ed25519'}, None)
```
simo5: don;t replace tests, keep the old ones and add new ones

jwcrypto/jwa.py
```diff
@@ -1297,6 +1306,14 @@ def instantiate_alg(cls, name, use=None):
         alg = cls.algorithms_registry[name]
         if use is not None and alg.algorithm_use != use:
             raise KeyError
+        if alg.status == 'prohibited':
+            raise InvalidJWAAlgorithm(
+                '%s is prohibited and must not be used' % name)
+        if alg.status == 'deprecated':
```
simo5: I am not sure we should warn in the EdDSA case, while the new RFC is all uppity about using fully qualified algorithms names, there is nothing particularly wrong with the old algorithm, and it will be in use for a long time before consumers will switch (if ever).

simo5 (changes requested):
1. We canNOT stop offering EdDSA, it will break all users
2. I am not even sure we can warn about deprecation, because the new and old algorithms are not *wire* compatible.
However if those warning come up only when someone intentionally check things rather than by default then it is probably ok.

EdDSA needs to remain enabled, so all the tests that assume it is disabled by default needs to be changed

jwcrypto/jwa.py
rjeffman: Makes sense, and actually explain the other points. I agree and will change the PR accordingly.

jwcrypto/tests.py
rjeffman: Fixed.

jwcrypto/jws.py
rjeffman: Fixed.

jwcrypto/jwa.py
rjeffman: Fixed.

rjeffman:
Modified implementation to allow opt-in deprecation warning or EdDSA algorithm. This will allow us to change behavior when non-fully specified algorithms are required, while allowing anyone that want to push adoption to have deprecation warnings.

simo5:
uhmm where did the support for the new algorithms go ?
all I see is the added warning for EdDSA now, and no new algorithms.

simo5:
Oh NVM I just see that the algorithms were already there ...

simo5 (approved):
LGTM

PR #390, opened 2026-08-11
Add initial support of Hybrid Public-Key Encryption

### LINKED ISSUE #387: Teak and implement HPKE once the RFC is published
https://datatracker.ietf.org/doc/draft-ietf-jose-hpke-encrypt/

rjeffman:
Hybrid Public-Key Encryption with JWE is in 'draft', and this PR implements the current RFC draft.

Due to pyca's cryptography limitations, X448 KEM and HPKE-[0:7] algorithms were not implemented as they are not currently exposed by the library, and I choose to not use private/protected methods/functions.

Also, PSK mode is not implemented as the cryptography HPKE implementation currently only supports `mode_base`.

See individual commits for more details on the implementation and decisions made.

### DIFF
--- a/jwcrypto/jwa.py
+++ b/jwcrypto/jwa.py
@@ -8,7 +8,7 @@ from binascii import hexlify, unhexlify
 
 from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.backends import default_backend
-from cryptography.hazmat.primitives import constant_time, hashes, hmac
+from cryptography.hazmat.primitives import constant_time, hashes, hmac, hpke
 from cryptography.hazmat.primitives.asymmetric import ec
 from cryptography.hazmat.primitives.asymmetric import padding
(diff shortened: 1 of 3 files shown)


simo5:
openssl only supports the base mode, so I guess that is why pyca also has that limitation

simo5 (changes requested):
Please add the relevant "Implements" comment line at the top of jwa.py 
Also move the official testvectors to the top of tests.py and identify their prvenance with a comment like for all other test vectors.

simo5:
This looks good to me, however maybe we should wait a little bit and see if https://github.com/pyca/cryptography/issues/15292 also gets resolved in the meanwhile ?

rjeffman:
I agree. @simo5 I removed the draft status just because you did an actual review.

IMHO, there's still some time for this to be merged (the RFC is still a draft).

simo5:
let's keep it as draft until we are ready to merge then.

PR #391, opened 2026-08-25
Reject non-empty encrypted key in bare ECDH-ES

simo5:
In direct key agreement mode (ECDH-ES), the encrypted key component must be empty. An allow_ek flag is introduced to enforce this requirement during unwrapping, raising an InvalidJWEData exception if a non-empty encrypted key is provided for bare ECDH-ES while still permitting it for ECDH-ES key wrap algorithms.

Assisted-by: Gemini:Gemini 3.7 Flash

Originally reported by @kieugiathinhphat


rjeffman (approved):
LGTM.

PR #392, opened 2026-08-26
Limit dot splits during JWE and JWS parsing

simo5:
When parsing compact JWE and JWS tokens, str.split('.') was previously executed without a maxsplit argument. Malformed tokens containing excessive dots could cause unnecessary memory allocation and processing overhead.

By specifying maxsplit parameters (5 for JWE and 3 for JWS), parsing efficiency is improved and tokens with too many components are quickly rejected as invalid.

Assisted-by: Gemini:Gemini 3.6 Flash


rjeffman (approved):
LGTM.

PR #395, opened 2026-08-31
Fix an issue with parsing headers with 'kid'

simo5:
This addresses an issue with JWS reported by @1321284827 and also brings up the related JWE feature to properly handle kid in per-recipient headers just like we support kid for multiple JWS signatures.

In both cases we had issues, but the JWE case was non-functional, while the JWS cases was problematic.


rjeffman (approved):
LGTM.

PR #397, opened 2026-09-14
Loop optimizations

rjeffman:
Optimize key_ops duplicate key search from O(n^2) to O(n) and audience claim matching from O(n*m) to average O(n + m).

Both changes increase memory use by O(n) by creating sets from lists.


simo5 (approved):
LGTM

PR #398, opened 2026-09-14
Avoid quadratic duplicate check on JWK key_ops

lissy93:
Hiya,
Short PR to fix a performance / security issue in the import_key function...

Previously, when jwcrypto loads a key, it checks the key_ops list for repeated values by comparing every single entry against every other one. Without any limit on how long the list can be. So time taken / CPU grows with the square of it's lengh.

So any service that decrypts ECDH-ES JWEs with an EC or OKP key, or that imports a JWK or JWK Set from untrusted input, can be stalled for minutes by a key with a very long key_ops list. Can result in a DoS, esp if the developer using this doesn't have their own validation or passes user-supplied JWK.

The fix here just replaces the nested loop with a single check what compares the list len with the length of its set. This makes loading linear, instead of quadratic.


simo5 (approved):
LGTM

PR #399, opened 2026-09-14
Simplify key_ops checks

simo5:
Restrict the number of key_ops accepted in input to the number of all possible Key Operations, more than that can't ever be assigned, so either there are duplicates or unsupported key_ops in the attribute.

Also simplify checking for key_ops/use compatibility by introducing an auxiliary dictionary.

Credit to @NotAFlightRisk for notifying this could be improved


NotAFlightRisk:
yay, thanks for the fix!! 🥳

rjeffman (approved):
LGTM.

PR #401, opened 2026-09-18
Make ML-DSA optional

rjeffman:
This patch detects if ML-DSA is usable during runtime, and configure jwcrypto to use it, if available. When not available, use of ML-DSA will raise a NotImplementedError.

This change allows the use of PyCA cryptography versions that do not support ML-DSA, which is required by some older systems.

Assisted-by: Claude Code Haiku 4.5 <noreply@anthropic.com>


simo5 (approved):
LGTM

